PR #14, opened 2015-09-08
add more matchers

### LINKED ISSUE #9: Implement default matchers
A few matchers that should be supported out of the box after #1 and return them from `require('testdouble').matchers`

## isA()

Should take a prototype and do whatever checking is necessary (beyond the buggy behavior of [instanceof](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) on primitive types) to ensure that something is-a.

```
when(testDouble(isA(Number), isA(String)).thenReturn("BOO")

testDouble(5, "hi") // "BOO"
```

# argThat

```
when(testDouble(argThat(function(a) { return a.b === 5})).thenReturn(88)

testDouble({b: 5}) // 88
testDouble({b: "not five"}) // undefined
```

## Other ideas

* `greaterThan`
* `lessThan`
* `startsWith`
* `endsWith`
* `contains` (for array elements as well as subsets of objects)

searls:
resolves #9 

This seems to cover all the obvious cases I can think of.

### DIFF
--- a/lib/matchers.coffee
+++ b/lib/matchers.coffee
@@ -34,4 +34,7 @@ module.exports =
         else
           throw new Error("the contains() matcher only supports strings, arrays, and plain objects")
 
+  argThat: (predicate) ->
+    __matches: (actual) ->
+      predicate(actual)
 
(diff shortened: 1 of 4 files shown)


searls:
merging b/c dictator

PR #19, opened 2015-09-10
Configure Testem

jasonkarns:
Presently, only the node launcher is configured. Browser testing still WIP

`npm start` runs testem (this would be the tdd dev flow)
`npm test` runs mocha (defaulting to spec output, but can be configured via .npmrc) and exits
`npm run test:debug` runs mocha and breaks into debug mode


searls:
:+1: LGTM

PR #25, opened 2015-09-14
Export `version` property on base object

jasonkarns:


searls:
:-1: on uglify for the main dep. I'm fine with an uglify second dep, but
since it's a test lib I'd rather leave it unminified in my own use

* Lets not commit to dist/ except when running `npm version` so only
released distributions ever hit git.

* can we get the version at the top of the built file in a comment? (e.g.
`/* testdouble@0.2.0 - https://github.com/testdouble/testdouble.js */`)

jasonkarns:
> on uglify for the main dep.

uglify is required to prune `package.json` out of the bundle. we need its unreachable-branch cleanup to keep package.json from being included from https://github.com/testdouble/testdouble.js/pull/25/files#diff-4ef89470f29f8c6989c09e54066b4623R7

> Lets not commit to dist/

I don't normally insert the built versions but did so intentionally in this branch so that we can see the uglified-without-version-property build and diff it with the uglified-with-version-property build. Can remove if still desired.

> version at the top of the built file in a comment?

we probably could simply by `cat`ing the string into the file, but browserify doesn't have any facilities for inserting comments. i only ever look for version strings from the browser, not the file itself. re-use uglify (proper) with the banner option?

searls:
> uglify is required to prune package.json out of the bundle. we need its unreachable-branch cleanup to keep package.json from being included from https://github.com/testdouble/testdouble.js/pull/25/files#diff-4ef89470f29f8c6989c09e54066b4623R7

Regardless, I don't want to minify it by default (I don't care if the package.json is in the built asset, but I also don't care if `testdouble.version` is defined and would rather exclude it entirely than have to minify it)

> Can remove if still desired. 

Yes I only want to have anything generated to `dist` or committed when I run `npm version` so that master is always guaranteed to point at a real release and I don't have to remember to reset it manually otherwise.

> we probably could simply by cating the string into the file,

That or any other means is fine. Bizarre to me that there's not a browserify functionality for this. Is publishing browser-consumed libraries not a common use case for browserify?

jasonkarns:
> Bizarre to me that there's not a browserify functionality for this. Is publishing browser-consumed libraries not a common use case for browserify?

Browser-consumed libraries are almost entirely consumed in minified/uglified form. (not to mention how impossible they are to read with browserify scaffolding) I think for that reason I'm seeing less and less use of comments for versioning and more use of the version property on the exported object instead. anecdotal

> I don't care if the package.json is in the built asset

If this is the case, then we can kill uglifyify and envify both, and just export `require('./package.json').version`. Much easier; only downside is having the package.json contents itself in the bundle. (there's actually a separate transform that's meant to do dead-code removal without the uglification part; which we could use in place of uglifyfy. except it's broken and doesn't actually work. huzzah)

jasonkarns:
removed all the dist builds and removed uglifyify/envify. so package.json contents are currently being included in the bundle. we can fix this without using uglifyify when https://github.com/zertosh/unreachable-branch-transform/issues/9 is fixed by using envify + unreachable-branch-transform

searls:
well this PR got a lot smaller.

i therefore demand we write a test for this since our coverage is pretty fab right now

jasonkarns:
@searls rebased to master. test added. put envify back in now that unreachable-branch-transform is usable in place of uglifyify.

So browser build runs through envify; gets version string from process.env.npm_package_version; runs through unreachable-branch-transform which kills the package.json require; so package.json is omitted from the build.

In node, the version is plucked from the require('package.json').version

searls:
LGTM

searls:
:sparkles:

PR #26, opened 2015-09-14
don't watch generated files

jasonkarns:
src_files are served *and* watched. but the src_files are generated JS
files, so each test run generates new js files, which are being watched
so it triggers another build, ad infinitum

testem supports serve_files which are served but not watched (which
requires watch_files to be specified separately)


searls:
:+1: did you test this out?

jasonkarns:
yeah, via numerous branches. (and i use this exclusively on my other projects for same reason)

searls:
LGTM :+1:

PR #30, opened 2015-10-01
Compile coffeescript

### LINKED ISSUE #29: Building CoffeeScript at runtime is slow.
Its impact is roughly 200ms per runtime.

For a baseline, in a file `td-repro.js` I have
```
console.log('hi')
```
when ran, it gives the output
```
$ time node td-repro.js
hi

real	0m0.107s
user	0m0.083s
sys	0m0.016s
```

after an `npm install testdouble`, I replace the contents of `td-repro.js` with
```
var td = require('testdouble');
var myTestDouble = td.create();

myTestDouble('foo', 5)
td.verify(myTestDouble('foo', 5))
console.log('success')
```
and run again, with these results:
```
$ time node td-repro.js
success

real	0m0.369s
user	0m0.371s
sys	0m0.035s
```

But, if I
- comment out the first line of `index.js` (rather `node_modules/testdouble/index.js`)
- and execute `coffee -c node_modules/testdouble/lib/*.coffee node_modules/testdouble/lib/store/*.coffee~`

then the times are *way* down:
```
$ time node td-repro.js
success

real	0m0.190s
user	0m0.164s
sys	0m0.026s
```

From a baseline of ~110ms per run, a library that adds 70-80ms per run is incredible compared to one that adds 250ms.

jasonkarns:
Resolves #29


searls:
Would you mind removing changes to dist/ in this PR. I only want those to change on version bump commits

package.json
```diff
@@ -8,40 +8,40 @@
     "email": "justin@testdouble.com",
     "url": "http://testdouble.com"
   },
-  "main": "index.js",
-  "browser": "lib/testdouble.coffee",
```
searls: no more "browser" entry?

searls:
Looks pretty exhaustive

package.json
jasonkarns: nope. the browser entry was only used to bypass the coffeescript registration being done in index.js. now that there is no coffeescript registration, the node entrypoint and browser entrypoint are the same.

searls: fantastic

jasonkarns:
The package.json diff looks bigger than it is.

- `clean` is the same as before except i moved it to the top (and swapped tmp for generated)
- `test` is the same except with long switch names instead of short flags
- `test:ci` is the same except reports all results in TAP
- `build` is the same except i use full switch names, and no longer register the .coffee extension


Bigger picture:

- coffee source is compiled before a browserify `build` so browserify doesn't do the coffee compilation itself (no more coffeeify)
- thus both node and browser entrypoints are the same, which is the compiled JS under `generated/`
- coffee source is compiled before `publish`
- thus the main entrypoint exists when published to registry as well as `npm install` with no args
- coffee tests *and* source are compiled to `generated/` before testem browser test runs thus browser tests point to `generated/`

mocha tests still point to coffee source, though this can be changed if desired.

The important thing to me is that the same coffee compiler is used to compile for distribution (`npm run compile`) as does the compilation under test. Mocha just delegates to the locally-installed coffee compiler so this holds. (It did *not* hold when browserify was using coffeeify, so coffeeify was ripped out.) We could go a step further and do a coffee compilation of source and tests prior to the mocha run, that way we know the tests are hitting the same compiled JS as will be distributed. As it stands now, it's using the same coffee compiler, but we're not guaranteed the same compiler settings.

jasonkarns:
In order to see what the published tarball will look like:
- ensure the working copy is clean
- `npm run clean`
- `npm pack`

the tarball should include:
- bower.json
- package.json
- LICENSE
- README
- dist/testdouble.js
- generated/**/*.js

searls:
+1

searls:
Is this ready to merge?

jasonkarns:
I think so, though it's been sufficiently purged from my memory

searls:
Mind pushing to a @next tag or something? Just added you as owner

jasonkarns:
sure thing

searls:
merging because nobody knows if this works.

PR #44, opened 2015-10-27
Adding reset() to the Readme.md

BrianGenisio:


searls:
:sparkles:

kenhowardpdx:
@searls I don't see this documentation in master. Can you verify that test doubles can be reset as a one-off? I'm not seeing this behavior.

searls:
Eek you're right. I must have misread the OP's patch. Resets are always global

searls:
@kenhowardpdx this teaches me to read from my phone -- no resets are always global. There is not a one-off reset mechanism.

kenhowardpdx:
@searls This is fine by me. I was looking for a good pattern and stumbled on this PR which made me wonder if there was a way to do this. Glad to know the current documentation is accurate!

Keep up the good work!

PR #47, opened 2015-11-13
add info about `verify` failure to the thrown Error

BaseCase:
I'm working on a plugin for testdouble.js over at https://github.com/BaseCase/testdouble-chai which wraps `verify()` to add custom assertions to `chai`. In order to display useful failure output in my wrapper, I've added an `info` object to the Error that is thrown when `verify` fails.

I totally understand if you want to close this due to not wanting to add data to Errors, not wanting to change testdouble.js for what should be a standalone plugin, etc.! I can most likely accomplish the same goal without these modifications, but it would get messier and more coupled to testdouble internals on the plugin end of things.

### DIFF
--- a/lib/verify.coffee
+++ b/lib/verify.coffee
@@ -8,7 +8,13 @@ module.exports = (__userDoesPretendInvocationHere__, config = {}) ->
     if callsStore.wasInvoked(last.testDouble, last.args, config)
       # Do nothing! We're verified! :-D
     else
-      throw new Error(unsatisfiedErrorMessage(last.testDouble, last.args, config))
+      err = new Error(unsatisfiedErrorMessage(last.testDouble, last.args, config))
+      err.info =
+        testDouble: last.testDouble
+        calls: callsStore.for last.testDouble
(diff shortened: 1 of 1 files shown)


searls:
0. No need to apologize for your work
0. I highly doubt there's any real risk to adding properties to an `Error` object; might be worth a google if there's a convention here or any potential conflicts. My inclination is this is fine as-is, but perhaps best left undocumented for now
1. Let me get the master build fixed so that the status is accurate
2. Would like some tests for this. Test coverage in the lib so far is pretty phenomenally high so I'd like to keep it up while I can
3. This is a bit of a stretch, but it strikes me that the info on these errors should basically be the same stuff you'd want to see on `td.explain`. What if this & explain each relied on an info() method that got all the info we would want?

BaseCase:
Wha...how did I miss `td.explain`?? It's right there in the README and everything. That actually does exactly what I need! I'm gonna close this PR because it'd be silly to reimplement that stuff inside `verify`.
:tada: :tada:

PR #49, opened 2015-11-14
Better docs

searls:
Start the long task of writing better docs.

- [x] Installation
    - [x] Node.js
    - [x] Browser
        - [x] Curl
        - [x] npm
        - [x] bower
    - [x] Configuring with test library
      - [x] Resetting testdouble.js
- [x] How to use test double
    - [x] The purpose of test doubles
        - [x] To test interactions in code whose primary responsibility is facilitating the collaboration of sevveral dependencies ("a collaborator")
        - [x] Not intended, necessarily, to fake out third party / integrations, the focus is to unit tests, specifically to support isolated TDD workflows (link to discovery testing)
    - [x] Getting Started
      - [x] Start with a subject, which needs a few deps (stub, intermediary, side effect)
      - [x] Write the test setting those up with true fakes
      - [x] Make the test pass
      - [x] Ample links to creation, stubbing, verifying
    - [x] All about creating
        - [x] Functions
            - [x] Anonymous
            - [x] Named
        - [x] Objects
            - [x] Arrays of function names
            - [x] Prototype mirroring
            - [x] Proxy (Anonymous or Named)
    - [x] All about stubbing
        - [x] Definition
        - [x] Simple stub example
        - [x] Single arg
        - [x] Multi arg
        - [x] return sequences
        - [x] why not unconditional stubbing?
        - [x] matchers
        - [x] supported options
    - [x] All about verifying
        - [x] Simple verification
        - [x] Single arg
        - [x] Multi arg
        - [x] Matchers
        - [x] using argument captors
        - [x] supported options
    - [x] All about td.replace (node.js only)
        - [x] Warning about node 0.10 incompatibility
        - [x] Example with function bags 
        - [x] example of functions
        - [x] example of prototype constructors
        - [x] explain why modules are not supported
        - [x] explain how quibble works
    - [x] custom matchers
    - [x] Debugging test doubles
    - [x] Plugins / Extensions
        - [x] testdouble-chai
        - [x] testdouble-jasmine
    - [x] Frequently asked questions
- [x] drop an outline like this one into the README and link back to all the doc pages


searls:
Thanks for the comment. I might make that change. In reality, a lot of times, as developers we're tasked with some implementation details baked into the requirements we're handed (especially relating to protocols and the interaction of our work product with other technologies). For that reason, POSTing in an example like this isn't necessarily an implementation detail of the thing under test, since it affects extrinsic behavior.

In cases like the above, I actually value a little bit of technical telegraphing in the test, because a vague business-ey description doesn't add much. (And to anticipate one reply, if an extrinsic change were ever made to this dependency such that it'd use a websocket instead of a POST, my workflow with test doubles would be to _delete the dependency and its entire sub-tree of units_, so I'm not too worried about comment rot)

hanneskaeufler:
Thanks for the elaboration! I definately agree on some technical telegraphing (love the expression :tada:). You'd likely also write some tests for the `saveProblem` method/class, and this is where the `POST` mention would be the most fitting I guess? This would leave the `generate` method and its test/comment entirely untouched from the change.

searls:
BOOM. Finally done.

PR #52, opened 2015-11-18
Swap .replace with browser-specific function

jasonkarns:
Swaps the replace function (in browser) for a function that just throws an
error. No separate node/browser test designations yet. That'll be a separate PR


searls:
Was super slow to reviewing this, my apologies!

PR #61, opened 2015-12-04
Rename src and scripts

searls:


package.json
```diff
@@ -8,30 +8,29 @@
...
-    "compile:tests": "coffee --output generated/test --compile test",
+    "compile:node": "coffee --output lib --compile src",
+    "compile:test": "coffee --output generated/test --compile test",
+    "compile:browser": "browserify . --standalone testdouble --outfile $npm_package_config_build_file --ignore 'quibble'",
+    "compile": "npm run compile:node && npm run compile:test && npm run compile:browser",
     "test": "mocha --ui mocha-given --reporter $npm_package_config_mocha_reporter --compilers coffee:coffee-script --recursive test/node-helper.coffee test/lib",
-    "test:ci": "npm run test --testdouble:mocha_reporter=tap && testem ci",
-    "test:examples": "cd examples/node && npm i && npm test && cd ../..",
+    "test:example": "cd examples/node && npm i && npm test && cd ../..",
+    "test:ci": "npm run test --testdouble:mocha_reporter=tap && testem ci && npm run test:example",
```
jasonkarns: wondering if we should look at making test:ci only be `testem ci` (ie, including the test and test:example in testem proper)

package.json
```diff
@@ -8,30 +8,29 @@
...
-    "clean": "rm -rf generated dist && git checkout -- dist",
+    "clean": "rm -rf generated dist lib && git checkout -- dist",
     "start": "testem",
-    "compile": "coffee --output generated --compile lib",
-    "compile:tests": "coffee --output generated/test --compile test",
+    "compile:node": "coffee --output lib --compile src",
+    "compile:test": "coffee --output generated/test --compile test",
+    "compile:browser": "browserify . --standalone testdouble --outfile $npm_package_config_build_file --ignore 'quibble'",
+    "compile": "npm run compile:node && npm run compile:test && npm run compile:browser",
     "test": "mocha --ui mocha-given --reporter $npm_package_config_mocha_reporter --compilers coffee:coffee-script --recursive test/node-helper.coffee test/lib",
```
jasonkarns: should we just point mocha at the generated/test and drop the coffeeify transform? (assuming pretest does compile:node/compile:test) Would eliminate the differing of coffee compilers between what's tested and what's distributed

searls: This is a good idea but I'm not going to have any time for tdjs until after the holidays; could you please create an issue?

PR #68, opened 2016-03-01
Add optional logging to #function

phedinkus:


src/function.coffee
```diff
@@ -3,15 +3,17 @@ store = require('./store')
...
       testDouble.toString = -> "[test double for \"#{name}\"]"
     else
       testDouble.toString = -> "[test double (unnamed)]"
 
-createTestDoubleFunction = ->
+createTestDoubleFunction = (name, options)->
   testDouble = (args...) ->
     calls.log(testDouble, args, this)
+    if options.log
+      console.log("Test Double Function '%s' called with: ", name, args)
```
searls: Style points, would you mind taking advantage of interpolation here?

`console.log("Test Double Function '#{name}' called with: ", args)`

searls:
A couple questions:

* Could you explain why this feature is valuable to you?
* There are currently half a dozen ways to create a test double function; is there a way to implement this such that logging could be enabled regardless of which API the user used? (e.g. `td.function`, `td.object`, `td.replace`?)

I'm reading between the lines quite a bit here, but I suspect it'd be less impactful on the API and as-if-not-more useful to add a debug mode to the library which can be configured globally (is `td.configure` a thing yet? I forget). That way, every interesting call, stub, verification and so forth can be logged optionally and unobtrusively without worrying about propagating options everywhere. 

If the reason for wanting a logging test double is to debug WTF is going on, I'd rather have a global debug mode than encourage people to look for a potentially hard-to-find and set option on the test doubles that they're concerned with.

Separately, a couple requests:

* Could you please revert the changes to `dist/`? We're careful to only commit `dist` at release time on master.
* Let's document in `docs/*creating*.md` before merging

searls:
Going to close this, to give myself a dramatic reason to talk with @phedinkus about this in person tomorrow

Issue #77, opened 2016-03-14
Quibble in the browser


randycoulman:
I'm trying to set up my project to run the tests in the browser for debugging purposes, using webpack-dev-server.  When I do so, I run into issues with quibble.  I note from https://github.com/testdouble/testdouble.js/commit/f694c82205d0b621a63dcc1be4e45c7fd3297261 that there's been some work done to guard against `require`ing quibble in the browser, but there are two more unguarded references to it.  One is in `reset.coffee`, and the other is in `replace/module.coffee`.  I believe the former reference is causing my issue.

I'm seeing the following error message:
```
ERROR in ./~/quibble/lib/quibble.js
Module not found: Error: Cannot resolve module 'module' in /Users/randy/src/scratch-it/skraphaug/node_modules/quibble/lib
 @ ./~/quibble/lib/quibble.js 9:11-28
```

But looking at the stack trace in my browser, it looks like the `require` of quibble in `reset.coffee` is the root problem.

I'll be happy to provide any additional information or settings you need.

searls:
To be clear, your issue is that when you use td.js in your browser, you're getting unrelated quibble errors, right? It's not the case that you're trying to make quibble work in the browser (which isn't supported), right?

searls:
If so, then this is definitely a bug. What I don't understand is why our test suite hasn't caught it

randycoulman:
Yes, correct.  I'm not even using the functionality required by quibble, but I can't get a clean webpack compile because of the above error.

I'm not sure why the test suite isn't catching it.  For me, the error is originating where I'm calling `td.reset()` in my global Mocha `afterEach()`.

searls:
I have an idea: could you add a skeleton webpack project to the examples/ directory with a failing npm test command so i can add that to the build?

randycoulman:
I've added the example; there's not much there, but it does illustrate the problem.  I note that the fix in https://github.com/testdouble/testdouble.js/commit/f694c82205d0b621a63dcc1be4e45c7fd3297261 doesn't actually fix the problem.  If I simply do `var td = require('testdouble');` in my test file (without calling `td.reset()`), I continue to get the same error on the command-line, but this time the root cause is in `replace/index.js` instead.

searls:
[...]

PR #78, opened 2016-03-15
Add webpack example

randycoulman:
This example has a basic webpack configuration that attempts to build a test bundle.  There’s also an `npm run test:debug` script to allow running the tests in the browser (at `localhost:8081`).

The tests can’t be run from the command line, but this example demonstrates the issue reported in #77.  Once that issue is resolved, the example should build correctly, and the tests should run in the browser.


searls:
Going to close this since it looks like there won't be custom config needed to use webpack, though it'd be nice to kill that warning.

PR #80, opened 2016-03-16
Improve support for webpack users

searls:
- [x] pare down @randycoulman's webpack example from #78 to something more minimal that also executes the tests in a browser
- [x] get the webpack tests passing
- [x] update references to get it into the build
- [x] kill the big warning.

for reference the big warning is:

```
$ $(npm bin)/webpack --resolve-modules
Hash: 6804c40ba67cfb4666d4
Version: webpack 1.12.14
Time: 443ms
         Asset    Size  Chunks             Chunk Names
test-bundle.js  453 kB       0  [emitted]  main
   [0] ./test/math-problem-test.js 1.1 kB {0} [built]
   [1] /Users/justin/code/testdouble/testdouble.js/dist/testdouble.js 438 kB {0} [built]

WARNING in /Users/justin/code/testdouble/testdouble.js/dist/testdouble.js
Critical dependencies:
8:475-482 This seems to be a pre-built javascript file. Though this is possible, it's not recommended. Try to require the original source to get better results.
```


randycoulman:
@searls, I haven't tried this yet, but this might be a way to get rid of the warning.

Try changing the webpack config file to something like this:

```
var path = require('path')

module.exports = {
  entry: './test/math-problem-test.js',
  module: {
    noParse: [
      /node_modules\/testdouble\//
    ]
  },
  output: {
    filename: 'test-bundle.js',
    path: './built'
  },
  resolve: {
    alias: {
      testdouble: 'testdouble/dist/testdouble.js'
    }
}
```

I culled this recipe from the discussion here: https://github.com/webpack/webpack/issues/304

searls:
Yeah I'm sure that end users could all roll their own aliases but that seems like far more trouble than necessary when all we need is to be able to indicate to webpack that this file needs to be replaced:
https://github.com/testdouble/testdouble.js/blob/master/package.json#L35-L37

I have to imagine webpack supports something like this if not exactly this but after wasting a couple hours on it tonight I'm quickly losing hope.

gaearon:
Webpack definitely supports `browser` field in `package.json`.

package.json
```diff
@@ -28,13 +28,15 @@
...
     "postversion": "git push && git push --tags && npm publish",
+    "version": "npm run version:write && npm run compile && git add dist src/version.coffee",
     "prepublish": "npm run compile"
   },
   "browser": {
-    "./lib/replace/module.js": "./lib/replace/module.browser.js"
+    "./lib/replace/module.js": "./lib/replace/module.browser.js",
+    "quibble": "./lib/quibble.browser.js"
   },
   "browserify": {
```
jasonkarns: with the version pulled from it's own module now, I believe we could eliminate envify and unreachable-branch-transform from the browserify config. IIRC, they were only used for dealing with the version.

searls: ok, i'll try nixxing them before i cut a release

davemo:
:+1:

PR #83, opened 2016-03-19
doc: thenDo and thenThrow are no longer unimplemented.

verdammelt:
Both `thenDo` and `thenThrow` have been implemented and documented. However there was a reference to them being unimplemented earlier in the file. Changed that to reference their description later in the file.


searls:
whoopsy

PR #84, opened 2016-03-19
Stringify objects better

searls:
Right now, a lot of logging in `explain` and `verify` will give up and print `[Object object]` sort of strings, when it'd be much rather to print them out. I forked and published [this module](https://github.com/searls/stringify-object-with-one-liners) to do most of the heavy lifting

- [x] give us a way to print objects better
- [x] make sure we plug it in everywhere we inspect objects for logging
- [x] test it
- [x] add a `td.matchers.create()` API, as the scheme for printing their names becomes more complicated
- [x] convert td's built-in matchers to call through to `td.matchers.create()`
- [x] pass the test on better matchers printing.
- [ ] document the matchers.create API
- [ ] contact @basecase and @BrianGenisio


searls:
Looks like it works. The previous system would blow up on circular objects (b/c it was using native JSON), would jam everything into one line, and would double-quote keys b/c JSON.

So the output of an argument changes from: `{"joe":5,"jill":[1,"2",3]}` to `{joe: 5, jill: [1, "2", 3]}`

searls:
I just realized that failed verifications that contain matchers are reallllllly ugly.

If I say:
```js
someTestDouble(55)

td.verify(someTestDouble(td.matchers.isA(String)))
```

I get this monstrosity:

``` js
Unsatisfied verification on test double.

  Wanted:
    - called with `({
  __matches: function (actual) {
          if (type === Number) {
            return _.isNumber(actual);
          } else if (type === String) {
            return _.isString(actual);
          } else if (type === Boolean) {
            return _.isBoolean(actual);
          } else {
            return actual instanceof type;
          }
        }
})`.

  But was actually called:
    - called with `(55)`.
```

I think we should probably figure out a custom toString for matcher args; maybe based on function name?

searls:
Looks like we're good to merge! The above monstrosity will now print much more nicely as:

``` coffee
"""
      Unsatisfied verification on test double.

        Wanted:
          - called with `(isA(String))`.

        But was actually called:
          - called with `(55)`.
"""
```

Huzzah!!

searls:
cc / @BrianGenisio & @basecase -- since I've chatted with both of you about custom matchers before, I wanted to give you the heads up that this new feature is available.

BaseCase:
@searls Nifty! I'll see about supporting this in testdouble-chai sometime soon. :cool:

searls:
@basecase I'm also thinking about adding some metadata to the errors that verify throws. I assume that you're catching the errors to map to chai assertions?

BaseCase:
@searls Right now it catches all `verify` errors and then uses `td.explain` to figure out what assertion message to print, but extra info right on the error object would totally be useful.

PR #87, opened 2016-03-22
replace: Use object[property] to check if property exists

### LINKED ISSUE #86: td.replace() fails for method on prototype
```js
var td = require('testdouble');

var Foo = function() {
  console.log('foo');
};

Foo.prototype.bar = function() {
  console.log('bar');
};

var foo = new Foo();

var bar = td.replace(foo, 'bar');

bar();
```

results in

```
Error: td.replace error: No "bar" property was found.
```

which is most likely related to the use of the `hasOwnProperty()` function to check if the property exists.

Turbo87:
This PR resolves #86

### DIFF
--- a/src/replace/property.coffee
+++ b/src/replace/property.coffee
@@ -13,7 +13,7 @@ module.exports = (object, property, manualReplacement) ->
   return fakeThing
 
 ensurePropertyExists = (object, property) ->
-  if !object.hasOwnProperty(property)
+  if !object[property]
     throw new Error("td.replace error: No \"#{property}\" property was found.")
 
 wrapIfNeeded = (fakeThing, realThing) ->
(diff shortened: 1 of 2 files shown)


searls:
The way this PR changes the behavior, if a property does exist but is falsey, then the error will be raised, which is a breaking change. 

The way I see it, we could either avoid making a breaking change by adding the new condition as an `||`, or embrace the breaking change and be more specific into what's allowed by type-checking the property and verifying it is either `_.isFunction` or `_.isObject`.

Turbo87:
@searls any chance of a new release in the near future?

searls:
Yes, sorry for the delay, I've got a few changes in flight that have held me back. I'll release by end of day tomorrow at the latest

searls:
Hey @Turbo87 this landed last night in 1.3.0, but there was a regression breaking 0.10.x users, so the latest is 1.3.1

Turbo87:
@searls awesome,  thanks!

PR #88, opened 2016-03-23
README: Replace GIFs with code blocks

Turbo87:
The GIFs are hard to comprehend and having proper code samples makes it much easier to copy-paste and play around with.


searls:
Sure. I waffled on this, because I want to show the code in motion, but the GIFs aren't accessible. I think the right solution is to do a few ~1m narrated screencasts instead.

PR #100, opened 2016-04-09
Adopt Scripty for scripts management

searls:
Went hog-wild (where most users won't need to) in order to dogfood [scripty](https://github.com/testdouble/scripty) as thoroughly as possible.


searls:
Closing as stale

PR #101, opened 2016-04-24
Add testdouble-timers as plugin to docs

kuy:
I added [testdouble-timers](https://github.com/kuy/testdouble-timers) to the docs. But I feel that it doesn't fit with the description because `testdouble-timers` doesn't bridge the `verify()` method. Please move my link if you also feel that and separate the section 😄


searls:
Good point @kuy I will try to adjust the language before i merge it

searls:
Merged! Thanks @kuy!

Issue #107, opened 2016-06-03
Constructor replacements will fail instanceof checks


searls:
See #103 for backstory, but the tl;dr is that a constructor replaced with `td.replace` module replacement will fail an `instanceof` test when compared to the actual type.

searls:
This is now an acceptance criteria of #166, closing to track it there.

Issue #103, opened 2016-04-28
td.replace breaks subsequent require calls


scottaj:
This code example demonstrates the problem. Since `CantMakeHeadsOrTailsOfItError` is require'd after doing a `td.replace`, it fails the `instanceof` check in the test. 

It seems unexpected that replacing a different module would have any effect on other modules require'd after it.

To be clear either removing the td.replace, or moving the require of the error to before the td.replace fixes the issue, so it is easy to work around.

```javascript
...
const MyLib = require('../index');
const MyDependency = td.replace('../dependency');
const CantMakeHeadsOrTailsOfItError = require('../cant-make-heads-or-tails-of-it-error');

describe('My Lib', function() {
  ...
  describe('#tail', function() {
    it('throws an error if the provided object is not an array', function () {
      try {
        myLib.tail(7);
        throw new Error('Did not get expected exception.');
      } catch (error) {
        expect(error).to.be.an.instanceOf(CantMakeHeadsOrTailsOfItError);
      }
    });
  });
});
```

Here is a runnable project with the full code example and failing test from above: https://github.com/scottaj/td-instanceof-example

searls:
There are three issues in the example above, and two of them will confound the third (which is a real issue)

1. If you're using the `td.replace` API, it needs to be inside a `beforeEach` or some kind of `before` hook that gets reset after each test, paired with `td.reset()` in an `afterEach` hook so that the state of the universe can be properly restored.

2. Any dependencies you replace need to be `td.replace()`'d before you actually call `require` on the subject under test; otherwise, the library has no way to know what dependencies to replace when the subject is loaded (and will instead load real ones). (e.g. `MyDependency = td.replace('../dependency')` should be before `MyLib = require('../index')`)

3. Instantiable function dependencies `replace()`'d with testdouble.js will fail `instanceof` checks.

Can you resolve the first two issues and retitle/redescribe the problem in those contexts? Otherwise I'll close and open a new one for it.

scottaj:
Sure, I'll update it.

searls:
Closing. Follow along in #107

PR #108, opened 2016-06-03
Pass instanceof checks for constructor replacement

searls:
Not sure how or whether the cure is worse than the illness in this case, but here's a test demonstrating the issue in #107 and #103


searls:
Why doesn't this test fail?

searls:
This test was insufficient to begin with, but its working is now an acceptance criteria of #166

PR #109, opened 2016-06-03
Buried matcher support

Schoonology:
Matchers nested within an Object or Array (a.k.a. "buried" matchers) argument are not handled correctly.
- [x] Regression test
- [ ] Fix

### DIFF
--- a/test/src/when-test.coffee
+++ b/test/src/when-test.coffee
@@ -39,6 +39,13 @@ describe 'when', ->
     Then -> @testDouble(44, 5) == undefined
     Then -> @testDouble(88, "five") == undefined
 
+  describe 'using buried matcher', ->
+    Given -> td.when(@testDouble({ key: td.matchers.isA(String) })).thenReturn("yay")
+    Then -> @testDouble({ key: "testytest" }) == "yay"
+    Then -> @testDouble({ key: true }) == undefined
+    Then -> @testDouble({}) == undefined
+    Then -> @testDouble(42) == undefined
+
   describe 'stubbing sequential returns', ->
     context 'a single stubbing', ->
       Given -> td.when(@testDouble()).thenReturn(10,9)


searls:
just a ping that this PR is lost but not forgotten <3

uhoh-itsmaciek:
Is it still not forgotten? =D

I can look at implementing the fix if you have pointers for where to start.

searls:
@uhoh-itsmaciek this is still outstanding and we'd definitely like to see it implemented such that the test passes. Be sure to rebase current master first!

uhoh-itsmaciek:
I've rebased and added failing regression tests for `.verify` (which should behave the same as `.when`, right?) [in my fork](https://github.com/testdouble/testdouble.js/compare/master...uhoh-itsmaciek:deep-matchers?expand=1). I've also renamed the functionality to "deep" instead of "buried"  because I think that's more common, but I'm happy to rename it back.

I've looked through `when.coffee`, but don't quite follow where the actual matcher checks take place--it looks like `store/calls` keeps a global stack of calls to verify and `when` consumes from that stack, but still not sure where the low-level logic lies. Is `args-match.coffee` where I should be looking?

searls:
1. Matchers should definitely work the same for `when` and `verify`, as symmetry is part of the contract we set with users about how td.js works
2. Agree, "deep" is more obvious and typical
3. Yes, `args-match.coffee` is (to the best of my recollection) the place to start looking

uhoh-itsmaciek:
I've updated my branch with what I think is a working version *except* for error messages handling. This is the failure I get:

```
  241 passing (426ms)
  1 failing

  1) .verify using deep matchers unsatisfied then td.verify(_this.testDouble({ value: td.matchers.isA(String):
     Error: expected 'Unsatisfied verification on test double.\n\n  Wanted:\n    - called with `({\n  value: {\n    __name: "isA(String)",\n    __matches: function (actualArg) {\n          return config.matches(matcherArgs, actualArg);\n        }\n  }\n})`.\n\n  But was actually called:\n    - called with `({value: 55})`.' to equal 'Unsatisfied verification on test double.\n\n  Wanted:\n    - called with `({ value: isA(String) })`.\n\n  But was actually called:\n    - called with `({ value: 55 })`.'
```

(also I apologize for my hideous and clumsy Coffeescript; I haven't worked much with it or JS for that matter)

I *think* I need to make changes to `src/stringify/anything.coffee` to do similar recursive handling, but I get seemingly unrelated failures when I try to change things there. Any ideas?

uhoh-itsmaciek:
Actually at this point, I'm just going to open my own PR; I can't update this one and I think it will be easier to discuss if I have code to point to directly.

searls:
Fixed by #203 & landed in testdouble@2.0.0-pre.4

PR #127, opened 2016-08-31
explain plan better in stubbing description

### LINKED ISSUE #126: td.explain should indicate thenCallback / thenResolve / thenReject from thenReturn
```javascript
var fn = td.func();
td.when(fn()).thenResolve(true);
console.log(td.explain(fn));
```

**Expected**
```javascript
{
  callCount: 0,
  calls: [],
  description: 'This test double has 1 stubbings and 0 invocations.\n\nStubbings:\n  - when called with `()`, then resolve `null`.',
  isTestDouble: true,
}
```

The important part is `then resolve 'null'`.

**Actual**
```javascript
{
  callCount: 0,
  calls: [],
  description: 'This test double has 1 stubbings and 0 invocations.\n\nStubbings:\n  - when called with `()`, then return `null`.',
  isTestDouble: true,
}
```

I can make a pull request if accepted.

Moeriki:
Fixes #126.

Feedback very welcome!

### DIFF
--- a/src/explain.coffee
+++ b/src/explain.coffee
@@ -32,7 +32,15 @@ testdoubleDescription = (testDouble, stubs, calls) ->
 stubbingDescription = (stubs) ->
   return "" if stubs.length == 0
   _.reduce stubs, (desc, stub) ->
-    desc + "\n  - when called with `(#{stringifyArgs(stub.args)})`, then return #{stringifyArgs(stub.stubbedValues, ", then ", "`")}."
+    plan = switch stub.config.plan
+      when 'thenCallback' then 'callback'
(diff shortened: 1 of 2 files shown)


searls:
Could you wrap the thenCallback args with ()? Otherwise looks good

Moeriki:
I changed `then callback 'null', '8'` to `then callback (null, 8)` for `returnCallback` plan.

searls:
For consistency sake could you also wrap the `()` in backticks like the other values? Once you have that I'm good to merge

Moeriki:
Makes sense. Updated.

searls:
Landed in 1.6.1

PR #132, opened 2016-09-18
Upgrade to lodash 4.15.0

searls:
**Not ready to merge**

- [x] fix 71 failing unit tests
- [x] get `npm run test:ci` to pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "quibble": "./lib/quibble.browser.js"
   },
   "dependencies": {
-    "lodash": "^3.10.1",
+    "lodash": "^4.15.0",
     "quibble": "^0.3.0",
     "stringify-object": "^2.4.0"
   },
--- a/src/args-match.coffee
+++ b/src/args-match.coffee
@@ -1,4 +1,7 @@
-_ = require('lodash')
+_ =
+  every: require('lodash/every')
+  isEqual: require('lodash/isEqual')
+  isFunction: require('lodash/isFunction')
 
 module.exports = (expectedArgs, actualArgs, config) ->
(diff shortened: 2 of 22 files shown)


searls:
Down to 39 failures…

searls:
down to 19 failures…

searls:
down to 0 unit test failures, moving on to `npm run test:ci`

searls:
ok,`npm run test:ci` is also passing, but the distribution went from 460KB to 576KB 😦 

hey @jasonkarns I'd love your opinion before clicking merge:

1. anything else i should check before merging this to be sure the 3->4 migration was safe

2. how much effort might it be to only require the lodash funcs we actually use?

searls:
Welp, I did the individual requires and was pretty shocked by the size difference

```
$ du -h dist/*
564K	dist/testdouble-big-lodash-4.js
200K	dist/testdouble-lodash-individual.js
452K	dist/testdouble.js
```

This PR bumped us from 452K to 564K, but by individually `require`'ing our lodash functions, it gets us back down to 200K, which seems like a definite good. AFAICT I'm not using any lodash functions dynamically.

searls:
This failure is due to some 0.12 network issue. Merging with reckless abandon

searls:
Landed in 1.7.0

jdalton:
Cool. For non-coffee projects you can use [babel-plugin-lodash](https://github.com/lodash/babel-plugin-lodash) to make cherry-picking easier. Then if using webpack [lodash-webpack-plugin](https://github.com/lodash/lodash-webpack-plugin) would cut the bundle size further.

PR #136, opened 2016-09-22
Remove warning of implementations of ES2015 proxy

mgryszko:
As for Oct 22nd 2016, proxies are supported in recent versions of all major browsers and in node 6.x. Safari 10 joined that club a couple of days ago.


searls:
:+1:

PR #137, opened 2016-09-26
fix test script in example

davemo:
I saw a `test` folder but no `spec` folder so just assumed this was a typo.


davemo:
nevermind, totally realized this is the reporter and not the folder  :(

searls:
Haha phew

PR #142, opened 2016-10-11
Support deep matchers

### LINKED ISSUE #160: Allow matchers inside objects
If I have a function that takes parameters as an options object, like so:

```js
// foo.js
module.exports = {
  function bar({a, b}) {
    doSomething() 
  }
}
```

I would like to be able to do this:

```js
td.replace(foo, 'bar')
td.when(
  foo.bar({
    a: td.matchers.isA(Date),
    b: 'hello world'
  })
).thenReturn('happy days')
```

uhoh-itsmaciek:
Adds support for deeply nested matcher expectations.

Expectations like this
- `{ value: isA(String), otherValue: 'foo' }`
- `[ isA(String), isA(Number) ]`
- `{ value: { nestedValue: isA(String) } }`

and combinations thereof are supported.

Respects `config.ignoreExtraArgs`.

See previous work in #109

Resolves #109 
Resolves #160


uhoh-itsmaciek:
* [ ] fix error message handling
* [ ] add test for array matching
* [ ] add test for deeply nested matching

searls:
Thanks for putting up with coffeescript to get started on this. I haven't looked at the code yet, but based on your earlier comment, I may make a PR against this PR to stylize the code you have so far before we tick off the other features.

However, I won't have time to look at this until Friday, so perhaps you can keep working on this until then. 

Thanks for contributing!

uhoh-itsmaciek:
Thanks! I've pushed some additional tests (everything except error messages still seems to work), and I've granted you write access to my branch, so you shouldn't need to make another PR--you should be able to edit this one directly by pushing to the `deep-matchers` branch in my fork (though if you can avoid force-pushing there, that'd be great--it will make it easier for me to follow your edits).

searls:
Hey @uhoh-itsmaciek -- I just wanted to chime in here and let you know I completely dropped the ball on reviewing / testing this and potentially merging it. This is still a problem and I want to incorporate it or another solution before 2.0

searls:
If you have the energy to rebase against master to resolve conflicts and retest, that'd be great, but it's also fine if not

searls:
Nevermind, the web UI let me resolve the very minor conflict myself.

uhoh-itsmaciek:
No problem, and thanks for the rebase. What's the next step?

searls:
Next step I think would be to do a sanity-check for me that #160 would work (I haven't looked closely at this yet) and get the build passing

searls:
Regarding the build break: I believe you may want `_.isPlainObject` instead of `isObject`? If you know you want `isObject`, then just add it to `src/utils/lodash-wrapper`

uhoh-itsmaciek:
I think you're right: I do want `_.isPlainObject` (I'm pushing the change), but that doesn't fix the build. I think the issue there is error message handling, and I don't follow how that's hooked into this process.

PR #144, opened 2016-10-12
Allow for multiple captor invocations

marchaos:
Allowing multiple invocations to a captor which will store all captured values in .values. .value still remain for backwards compatibility, and will always contain the last invoked capture value.

### DIFF
--- a/src/matchers/captor.coffee
+++ b/src/matchers/captor.coffee
@@ -5,9 +5,7 @@ module.exports = ->
     capture: create
       name: 'captor.capture'
       matches: (matcherArgs, actual) ->
-        if not captor.values
-            captor.values = []
+        captor.values ||= []
         captor.values.push actual
         captor.value = actual
         return true
-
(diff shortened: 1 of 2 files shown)


src/matchers/captor.coffee
```diff
@@ -5,5 +5,9 @@ module.exports = ->
     capture: create
       name: 'captor.capture'
       matches: (matcherArgs, actual) ->
+        if not captor.values
```
searls: try `captor.values ||= []` and see if that works for you here to save a line

marchaos:
Updated.

searls:
Looks good! Thanks!

searls:
Can you open a PR that documents this in the docs/ section on captors? Once we have that I'll cut a release

marchaos:
Sure! - https://github.com/testdouble/testdouble.js/pull/145
Let me know if that's the best place.

marchaos:
Probably easier to preview here - https://github.com/marchaos/testdouble.js/blob/master/docs/6-verifying-invocations.md#capturing-multiple-invocations-with-tdmatcherscaptor

PR #150, opened 2016-10-24
Gives better hint when verification fails while ignoring extra args

### LINKED ISSUE #60: Note when `ignoreExtraArgs` is used for a failed verification
Suppose you have this invocation and failed verification:

``` javascript
print('some', 'stuff', 'out', 'like', 8)

td.verify(print('some', 'stuff', 'NOPE'), {ignoreExtraArgs: true})
```

This will throw:

```
Error: Unsatisfied verification on test double.

  Wanted:
    - called with `("some", "stuff", "NOPE")`.

  But was actually called:
    - called with `("some", "stuff", "out", "like", 8)`.
```

It would be better if it printed a hint like:

```
  Wanted:
    - called with `("some", "stuff", "NOPE")`, ignoring any additional arguments.
```

samjonester:
This closes #60.

### DIFF
--- a/src/verify.coffee
+++ b/src/verify.coffee
@@ -38,7 +38,7 @@ unsatisfiedErrorMessage = (testDouble, args, config) ->
   Unsatisfied verification on test double#{stringifyName(testDouble)}.
 
     Wanted:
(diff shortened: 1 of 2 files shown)


searls:
Yay!

PR #151, opened 2016-10-26
Fakes properties tacked onto default exported function

### LINKED ISSUE #99: td.replace for functions that have additional function properties
It's not uncommon for a node module to export a function but to also tack on some additional functions as properties on the function.

Right now, `td.replace` does not handle this well. It just replaces the top-level function. It should probably look for these additional functions and replace those, as well.

Feels a little heavy-handed / demetertastic though. Interested in feedback.

samjonester:
Replaces imitatable enumerable properties on an exported function.
Does not replace inherited properties.
Allows expected replacement for module like:

``` javascript
function foo() {}
foo.bar = function() {}
module.exports = foo
```

Closes #99

Although not explicitely tested, this will also work for a module like:

``` javascript
module.exports = function foo() {}
modue.exports.bar = function() {}
```

### DIFF
--- a/src/replace/module.coffee
+++ b/src/replace/module.coffee
@@ -1,14 +1,15 @@
 quibble = require('quibble')
 imitate = require('./imitate')
 wrapIfNeeded = require('./wrap-if-needed')
+imitateProperties = require('./imitate-properties')
 
 quibble.ignoreCallsFromThisFile()
+_ =
+  merge: require('lodash/merge')
 
 module.exports = (path, stub) ->
   return quibble(path, stub) if arguments.length > 1
   realThing = require(quibble.absolutify(path))
   fakeThing = imitate(realThing, path)
   quibble(path, wrapIfNeeded(fakeThing, realThing))
-  return fakeThing
-
-
+  return _.merge(fakeThing, imitateProperties(realThing))
--- a/src/replace/imitate-properties.coffee
+++ b/src/replace/imitate-properties.coffee
@@ -1,0 +1,21 @@
+wrapIfNeeded = require('./wrap-if-needed')
+imitate = require('./imitate')
+_ =
+  isFunction: require('lodash/isFunction')
+  keys: require('lodash/keysIn')
+  reduce: require('lodash/reduce')
+  set: require('lodash/set')
+
+module.exports = (realThing) ->
+  return {} unless _.isFunction(realThing)
+  _.reduce imitatableKeys(realThing), wrapKey(realThing), {}
+
+imitatableKeys = (realThing) ->
+  _.keys(realThing).filter (key) ->
+    _.isFunction(realThing[key])
+
+wrapKey = (realThing) -> (newThing, key) ->
+  _.set(newThing, key, wrapIfNeeded(
+    imitate(realThing[key], '.'+key),
+    realThing[key]
+  ))
--- a/test/fixtures/car.coffee
+++ b/test/fixtures/car.coffee
@@ -6,3 +6,4 @@ module.exports =
   turn: require('./turn')
   brake: require('./brake')
   lights: require('./lights')
+  shift: require('./shift')
--- a/test/fixtures/shift.js
+++ b/test/fixtures/shift.js
@@ -1,0 +1,9 @@
+function shift() {
+  return 'Faster'
+}
+
+shift.neutral = function() {
+  return 'Coast'
+}
+
+module.exports = shift
--- a/test/src/replace/index-test.coffee
+++ b/test/src/replace/index-test.coffee
@@ -150,6 +150,7 @@ describe 'td.replace', ->
     Given -> @passenger = td.replace('../../fixtures/passenger') #<-- a constructor func
     Given -> @honk = td.replace('../../fixtures/honk') #<-- a plain ol' func
     Given -> @turn = td.replace('../../fixtures/turn') #<-- a named func
+    Given -> @shift = td.replace('../../fixtures/shift') #<-- a named func with property DIRECTION
     Given -> @brake = td.replace('../../fixtures/brake', 'ANYTHING I WANT') #<-- a manual stub bc brake does not exist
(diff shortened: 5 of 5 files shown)


src/replace/module.coffee
```diff
@@ -2,6 +2,9 @@ quibble = require('quibble')
 imitate = require('./imitate')
 isConstructor = require('./is-constructor')
 wrapWithConstructor = require('./wrap-with-constructor')
+_ = {
+  keys: require('lodash/keys')
```
samjonester: This ignores inherited properties. Not sure if this should be the desired functionality.

src/replace/imitate.coffee
```diff
@@ -16,3 +16,6 @@ module.exports = (realThing, optionalName) ->
     tdFunction(if realThing?.name then realThing.name else optionalName)
   else
     log.error("td.replace", "\"#{optionalName}\" property was found, but test double only knows how to replace functions, constructors, & objects containing functions (its value was #{stringifyAnything(realThing)}).")
+
+module.exports.canImitate = (realThing) ->
```
samjonester: The other option here is to accept an arg that would skip erroring on demand, but logic would still be required to avoid adding empty imitations to the fakeThing.

src/replace/module.coffee
```diff
@@ -16,4 +19,16 @@ wrapIfNeeded = (fakeThing, realThing) ->
   if isConstructor(realThing)
     wrapWithConstructor(fakeThing)
   else
+    replaceProperties(fakeThing, realThing)
```
samjonester: Not sure if this is the proper place for this action, or if it introduces temporal coupling.

src/replace/imitate.coffee
```diff
@@ -10,9 +10,18 @@ wrapWithConstructor = require('./wrap-with-constructor')
 stringifyAnything = require('../stringify/anything')
 
 module.exports = (realThing, optionalName) ->
+  if canImitate(realThing)
+    imitate(realThing, optionalName)
+  else
+    log.error("td.replace", "\"#{optionalName}\" property was found, but test double only knows how to replace functions, constructors, & objects containing functions (its value was #{stringifyAnything(realThing)}).")
+
+module.exports.canImitate = canImitate
```
samjonester: The other option here is to accept an arg that would skip erroring on demand, but logic would still be required to avoid adding empty imitations to the fakeThing.

src/replace/module.coffee
```diff
@@ -10,10 +14,24 @@ module.exports = (path, stub) ->
...
-  return fakeThing
+  return _.merge(fakeThing, imitateProperties(realThing))
 
 wrapIfNeeded = (fakeThing, realThing) ->
   if isConstructor(realThing)
     wrapWithConstructor(fakeThing)
   else
     fakeThing
+
+imitateProperties = (realThing) ->
```
samjonester: This function is ugly. Mostly because I haven't done much coffee script :/

src/replace/module.coffee
```diff
@@ -10,10 +16,24 @@ module.exports = (path, stub) ->
...
+  return {} unless _.isFunction(realThing)
+  _.reduce imitatableKeys(realThing), wrapKey(realThing), {}
+
+imitatableKeys = (realThing) ->
+  _.keys(realThing).filter (key) ->
```
samjonester: This ignores inherited properties. Not sure if this should be the desired functionality.

jasonkarns: Drop in [`_.keysIn`](https://lodash.com/docs/4.17.0#keysIn) instead if inherited keys are desired, as well.

I'll leave the "desirability" determination up to others, since I don't have any context around this.

searls:
just a ping to say i have not forgotten about this PR and hope to look at it this week

searls:
Hey Sam, sorry for the merge pain I caused you but wanted to let you know this is still a necessary patch, as demonstrated by this REPL

```
Let's play with testdouble.js!  💚
td > var dog = { woof: function() {} }; 
td > dog.woof.bark = function() {}
[Function]
td > td.replace(dog, 'woof')
td > dog.woof
td > dog.woof.bark
```

samjonester:
That's all right. I'll get it resolved. Do you have a solid opinion about this? https://github.com/testdouble/testdouble.js/pull/151#discussion_r89205593

searls:
OK I weighed in

samjonester:
Conflict resolved. I've updated to `keysIn` based on your comment as well. That reasoning makes sense to me. I wasn't able to understand how they would function differently in this context.

searls:
I can vouch for the reasoning but not how it actually works. Can you add a test on an actually-inherited thing if you haven't already?

samjonester:
I experimented with this, and `keysIn` isn't necessary. I've amended the PR to remove it.

`keysIn` will retrieve properties that are added via `prototype`, and therefore only exist on constructor functions that will be instantiated via `new`. The properties of constructor functions are not replaced via this flow. I've come up with an (IMO very contrived) example, that functions the same with `keys` as it does with `keysIn` because of this fact.

``` coffee
# car.coffee
module.exports =
  # ...
  cdPlayer: require('./cd-player')
```

``` js
// cd-player.js
function Radio() {}
Radio.prototype.tuneIn = function(station) {
  return 'Playing your favorites on ' + station
}

function cdPlayer() {
  return 'Loud Noises!'
}

cdPlayer.radio = Radio

module.exports = cdPlayer
```

``` coffee
# index-test.coffee
describe 'faking inherited properties with prototypal values', ->
      Given -> @cdPlayer = td.replace('../../fixtures/cd-player')
      Given -> @car = require('../../fixtures/car')
      Given -> @radio = new @car.cdPlayer.radio()
      When -> td.when(@radio.tuneIn('Hot 101')).thenReturn('static!')
      When -> @result = @radio.tuneIn('Hot 101')
      Then -> 'static!' == @result
```

searls:
Thanks @samjonester -- I'm too mentally fried to think through this today but let's find a time to pair on this and other tdjs issues early next week while I'm on break

PR #153, opened 2016-11-04
Add failing test for resetting replaced modules.

Schoonology:
While not as useful as a fix patch, this is a (currently) failing test that shows how I (currently) expect `replace` and `reset` to work together. Let me know if either is wrong.


searls:
Was this tracked by an issue? This sounds really bad.

Schoonology:
I don't think I created an issue first, sorry. I think the only pre-PR chatter was over Slack.

searls:
This test was passing so I'm just going to close it b/c we got enough going on.

PR #155, opened 2016-11-22
Changes real call summary to display times called

### LINKED ISSUE #85: Improve output for `times` verification failures
See this test:

``` coffee
      context '1 time, unsatisfied (with 2)', ->
        When -> @testDouble()
        And -> @testDouble()
        Then -> shouldThrow (=> td.verify(@testDouble(), times: 1)), """
          Unsatisfied verification on test double.

            Wanted:
              - called with `()` 1 time.

            But was actually called:
              - called with `()`.
              - called with `()`.
          """
```

I'd rather see it end with:

``` coffee
          """
            But was actually called:
              - called with `()` 2 times.
          """
```

This would require some care to make sure we account for matchers of different types of calls when we compare and print the actuals (since if the matchers satisfy they'll 'look' the same to the lib. We'd need a way to disable matchers for this purpose)

samjonester:
This is a starting place to close #85. There is something that this solution does not account for.

What should happen in a situation where the use wants to verify that a test double is called 2 times, but it was called 3 times with different arguments?

### DIFF
--- a/src/verify.coffee
+++ b/src/verify.coffee
@@ -39,17 +39,19 @@ unsatisfiedErrorMessage = (testDouble, args, config) ->
 
     Wanted:
       - called with `(#{stringifyArgs(args)})`#{timesMessage(config)}#{ignoreMessage(config)}.
-  """ + invocationSummary(testDouble)
+  """ + invocationSummary(testDouble, config)
 
 stringifyName = (testDouble) ->
   if name = store.for(testDouble).name
     " `#{name}`"
   else
     ""
-invocationSummary = (testDouble) ->
+invocationSummary = (testDouble, config) ->
   calls = callsStore.for(testDouble)
   if calls.length == 0
     "\n\n  But there were no invocations of the test double."
+  else if config.times? && config.times != calls.length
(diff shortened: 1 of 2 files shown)


searls:
I don't think this quite solves the issue because it doesn't yet group identical args entries, so if there's a call on the double that doesn't match, then it gets unrolled. 

Demo:

```
td > foo = td.func()
td > foo(); foo(); foo(3); foo();
undefined
td > td.verify(foo(), {times: 4})
Error: Unsatisfied verification on test double.

  Wanted:
    - called with `()` 4 times.

  But was actually called:
    - called with `()`.
    - called with `()`.
    - called with `(3)`.
    - called with `()`.
```

searls:
Hey Sam, I spent all night reflecting on this issue and how it maybe ougtht to be, so  I just proposed this as an improvement https://github.com/testdouble/testdouble.js/issues/161

samjonester:
Thanks for taking a look. I agree there's room for improvement and this solution doesn't handle the situation you proposed well. Would you like to close this PR + issue since you've created #161? I've still got the branch with this work locally for reference, and would like to take a stab at fixing #161.

searls:
Closing in favor of bigger #161 initiative

PR #158, opened 2016-12-10
Sort out replacement of especially 6+ Classes

### LINKED ISSUE #119: td.replace doesn't properly detect ES6 classes
When using Node 6 with ES6 classes the `td.replace` call doesn't properly detect ES6 classes as constructors because the methods of ES6 classes are not enumerable. The constructor is not wrapped as expected.

The affected line is https://github.com/testdouble/testdouble.js/blob/65251fe625574aea0569140768f993641d53620f/src/replace/is-constructor.coffee#L5

Test case to reproduce:

```
# index.js
const td = require('testdouble');
const MyES5Class = td.replace('./MyES5Class');
const MyES6Class = td.replace('./MyES6Class');

td.when(MyES5Class.mymethod()).thenReturn(5); // <-- works
td.when(MyES6Class.mymethod()).thenReturn(6); // <-- MyES6Class.mymethod is not a function

# MyES5Class.js
function MyES5Class() {}
MyES5Class.prototype.mymethod = function () {
  return 0;
};

module.exports = MyES5Class;

# MyES6Class.js
class MyES6Class {
  constructor() {}
  mymethod() {
    return 0;
  }
}

module.exports = MyES6Class;
```

### LINKED ISSUE #157: Mocking a Class
Thanks for creating such an awesome library! We've been enjoying using it so far.

Is it possible to mock a class with testdouble.js? For instance, let's say I wanted to mock the following module:

``` js
module.exports = class ES6Class {
  test() {
    return "hello";
  }
}
```

I'd like to mock this class like so:

``` js
let ES6ClassMock = testdouble.replace('./es6_class');
testdouble.when(ES6ClassMock.prototype.test()).thenReturn("This doesn't work :(");
```

When I run this, I got the following error:

```
TypeError: ES6ClassMock.prototype.test is not a function
```

When I try this with an ES5 constructor function, I receive a different error:

``` js
function ES5Class() {}

ES5Class.prototype.test = function() {
  return "hello";
};

module.exports = ES5Class;
```

``` js
let ES5ClassMock = testdouble.replace('./es5_class');
testdouble.when(ES5ClassMock.prototype.test()).thenReturn("This doesn't work :(");
```

```
TypeError: Cannot read property 'test' of undefined
```

Is it possible to accomplish what I'm trying to do? Thanks in advance!

searls:
- [x] Make native ES class replacement work (node 6 & up
- [x] Make babel'd ES class replacement work
- [x] Make `default export class` work (requires changing function bag behavior)
- [x] Make func bag improvements work for property replacement as well as module replacement
- [x] fix 0.10 failures
- [x] make it not a mess
- [x] add explicit tests
- [x] test against node < 6
- [x] document it

Resolves #119
Resolves #157


searls:
Landed in 1.1.0

Issue #167, opened 2016-12-16
typings for typescript


mkusher:
Do you have any plans of creating typings or will your accept pr with typings?

searls:
I don't know typescript but I'd be happy to improve support for it! Especially because test doubles are much more valuable with type systems!

mkusher:
@searls nice, then I'll try to create some working definition file on the weekend

PR #169, opened 2016-12-17
Typings for typescript

mkusher:
Resolves https://github.com/testdouble/testdouble.js/issues/167
I'm not sure about how to test it properly in your configuration, but it looks like my variant would be ok.
I've added feature from new ts 2.1, but it's only needed for one case `td.object(['a', 'b'])`. Also not sure what to add into README.md.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,8 +1,6 @@
 export type TestDouble = Function;
 
-export type DoubledObjectWithKey<Key extends string> = {
-  [P in Key]: TestDouble;
-};
+export type DoubledObjectWithKey<Key extends string> = {};
 
 export type DoubledObject<Subject> = Subject;
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -40,3 +40,28 @@ export type Matchers = {
 };
 
 export const matchers: Matchers;
+
+export function replace(path: string, f?: any): void;
+export function replace(path: {}, property: string, f?: any): void;
+
+export function reset(): void;
+
+export type VerificationConfig = {
+  ignoreExtraArgs?: boolean;
+  times?: number;
+};
+
+export function verify(a: any, check?: VerificationConfig): void;
+
+type Call = {
+  context: {};
+  args: any[];
+};
+
+export type Explanation = {
+  callCount: number;
+  calls: Call[];
+  description: string;
+}
+
+export function explain(f: TestDouble): Explanation;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,42 @@
+export type TestDouble = Function;
+
+export type DoubledObjectWithKey<Key extends string> = {
+  [P in Key]: TestDouble;
+};
+
+export type DoubledObject<Subject> = Subject;
+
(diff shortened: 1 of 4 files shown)


orazyto:
@mkusher - thanks for putting this together!
I've tried it on my machine - td is now recognized in my Typescript spec (I can get an example working).

When I run my tests, I do the following errors though:
ERROR in [default] /Users/orazyto/Documents/RapidPrototype/test_project/node_modules/testdouble/index.d.ts:4:2 
A computed property name in a type literal must directly refer to a built-in symbol.

ERROR in [default] /Users/orazyto/Documents/RapidPrototype/test_project/node_modules/testdouble/index.d.ts:4:2 
A computed property name must be of type 'string', 'number', 'symbol', or 'any'.

ERROR in [default] /Users/orazyto/Documents/RapidPrototype/test_project/node_modules/testdouble/index.d.ts:4:3 
Cannot find name 'P'.

ERROR in [default] /Users/orazyto/Documents/RapidPrototype/test_project/node_modules/testdouble/index.d.ts:4:8 
Cannot find name 'Key'.

mkusher:
@orazyto you are using typescript <= 2.1

orazyto:
@mkusher - you are right, my mistake. 
Since we're using Angular2, upgrading to typescript 2.1 seems to be currently blocked - as described here:
http://stackoverflow.com/questions/40309510/typescript-2-1-coupled-with-the-angular-2-compiler-silently-fails-to-compile-pro
https://github.com/angular/angular/pull/13294
I'm awaiting this fix to be able to move forward testing your typings file.

mkusher:
@orazyto oh, I see. Well, I've removed this mapped type for arrays, they now are untyped. Can you check the branch now?

orazyto:
@mkusher - Thanks! It's working now with Typescript 2.0 in my Angular2 project :-)

mkusher:
@orazyto ;)

searls:
Two questions:

1. Is there anything I need to know about this thing as a maintainer other than having this in the repo / npm bundle?

2. Can we move this typescript config json outside of `src/`? I want to keep `src` for the actual code in the library and it seems like a config file wouldn't belong there. I'm fine with a dot-file in the root or a `config/typescript.json`. Let me know if this is counter-idiomatic or wouldn't work

mkusher:
@searls 

1. It should be included in npm package(not be excluded by npmignore or something similar). It describes package's public api, so if public api is changed the d.ts file needs to be changed
2. I can put index.d.ts anywhere you want, the path to index.d.ts is specified in package.json

searls:
Is `index.d.ts` the conventional location? Seems like people know to look for it or something like it when they glance at a repo

mkusher:
@searls from documentation:
```
Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so.
```

searls:
neato, looks good

searls:
Landed in 1.10.1

PR #174, opened 2016-12-30
Captor typings

danielmoore:
This PR:

1. Changes all type aliases to multi-property objects to interfaces.
    - This is a backwards compatible change; it will not affect compilation.
    - Interfaces allow [declaration merging], while type aliases do not. So, if the type definitions fall behind the implementation (for instance, missing the `captor()` function), downstream users can declaration-merge in a patch to keep going until the official type def is updated.
1. Adds `td.matchers.captor()` to the type definitions.

[declaration merging]: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -16,7 +16,7 @@ export function object<Key extends string>(props: Key[]): DoubledObjectWithKey<K
 // When passed general object
 export function object<T>(object: T): DoubledObject<T>;
 
-export type Stubber = {
+export interface Stubber {
   thenReturn(...args: any[]): TestDouble;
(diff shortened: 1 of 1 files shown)


searls:
:+1: Going to just blindly accept all typescript PRs until someone tells me otherwise, since I don't understand them well enough to verify myself.

danielmoore:
@searls if you'd like some help reviewing your TypeScript PRs, I'd be happy to pitch in

searls:
Thanks @danielmoore I appreciate that.

searls:
Landed in 1.10.2

PR #175, opened 2017-01-01
Make dev possible on windows

searls:
- [X] Remove config variables because env expansion can't be done with both $ and %% style prompts
- [X] Upgrade testem to blow up less on Windows
- [X] Convert testem.js to testem.json since it is no longer dynamic
- [X] Fix the browser-test helper to define global. (how was this working before?)
- [X] Fix the testem config to remove lodash (how was this working before?)
- [ ] get the example projects to build (they currently use symlinks to reference in-repo module and that's failing)
- [ ] See `npm run test:ci` pass on a windows box
- [ ]  Verify that npm run test:cover:run still works, w/ the env setting the way it is, it won't run on windows but doesn't really need to
- [ ] set up an appveyor build
- [ ] publish a patch from a windows dev box


package.json
```diff
@@ -9,20 +9,17 @@
...
-    "mocha_reporter": "dot"
-  },
   "scripts": {
     "clean:dist": "git checkout -- dist",
-    "clean": "rm -rf generated dist lib .coverage && npm run clean:dist",
+    "clean": "rimraf generated dist lib .coverage && npm run clean:dist",
     "compile:node": "coffee --output lib --compile src",
     "compile:test": "coffee --output generated/test --compile test",
-    "compile:browser": "browserify . --standalone td --outfile $npm_package_config_build_file --ignore 'quibble' -p headerify",
+    "compile:browser": "browserify . --standalone td --outfile dist/testdouble.js --ignore 'quibble' -p headerify",
```
jasonkarns, line 17:
```suggestion
    "compile:browser": "browserify . --standalone td --outfile dist/testdouble.js --ignore quibble -p headerify",
```

package.json
```diff
@@ -9,20 +9,17 @@
...
     "compile": "npm run compile:node && npm run compile:test && npm run compile:browser",
-    "test": "mocha --ui mocha-given --reporter $npm_package_config_mocha_reporter --compilers coffee:coffee-script $EXTRA_MOCHA_ARGS --recursive test/node-helper.coffee test/src",
+    "test": "npm run test:base -- --recursive test/node-helper.coffee test/src",
+    "test:base": "mocha --ui mocha-given --reporter dot --compilers coffee:coffee-script",
     "test:cover:instrument": "istanbul instrument lib -o .coverage/lib",
-    "test:cover:run": "EXTRA_MOCHA_ARGS=\"--reporter mocha-istanbul\" COVER=.coverage/lib ISTANBUL_REPORT_DIR=.coverage ISTANBUL_REPORTERS=lcov,html npm run test",
+    "test:cover:run": "COVER=.coverage/lib ISTANBUL_REPORT_DIR=.coverage ISTANBUL_REPORTERS=lcov,html npm run test:base -- --reporter mocha-istanbul --recursive test/node-helper.coffee test/src",
```
jasonkarns, line 22:
```suggestion
    "test:cover:run": "npm run test:base -- --reporter mocha-istanbul --recursive test/node-helper.coffee test/src",
```

searls:
Marking as stale. WLS is advanced enough now that I imagine most people would just use that (I'm also imagining it working)

PR #177, opened 2017-01-17
chore: add yarn.lock

JaKXz:
Hi! 

I had just cloned the repo to get ready to make a PR for a feature request and wondered why on post-install I already had diffs in `dist/testdouble.js`. As it turns out, I have pulled a newer version of `coffee-script` than the last version was published with, and I figured I might as well suggest adding a `yarn.lock` to the repo.

Let me know what you think. I have been using `yarn` in my OSS, private, and company projects (including those on top of the rails stack) and having a Gemfile.lock equivalent for client code is very nice [on top of the speed improvements etc].


searls:
So far I haven't added any yarn to my OSS because it requires Node 4 and most of my libraries support back to Node 0.10. When I drop older support I'll add yarn. 

Thanks!

JaKXz:
@searls fair enough! When are you planning to do that, when they just break? LTS support has been dropped for a few months on 0.10 as you probably know.

searls:
Most testing libs in most languages are pretty conservative about dropping support for old versions because lots of people in the wild depend on very old versions and when they need to upgrade to a newer version, they tend to depend on testing.

I haven't decided when to drop support at this time

JaKXz:
FWIW travis [supports yarn and npm living alongside each other in harmony](https://travis-ci.org/testdouble/testdouble.js/jobs/192561496) so adding the lockfile isn't actually going to hurt anyone AFAICT. Maybe I'm missing something in the build /bundle step of this library but I assume that's covered by the tests which are all green?

JaKXz:
@searls I also just noticed that node v0.10 is technically "dropped" in the engines section:

https://github.com/testdouble/testdouble.js/blob/master/package.json#L89

Food for thought, perhaps testdouble v2? :)

searls:
Yeah, the reason we did that is because 0.10 support was only partial because 0.10 cannot support Quibble/td.replace() API

PR #178, opened 2017-01-17
feat: contains allows for regexps against strings

JaKXz:
This is my initial stab at adding RegExp support to the contains matcher. Let me know if/what kind of [other] edge cases you'd like me to cover :)

Also if this is not considered to be a good idea, I'd love to know why that decision was made. If I understand the `argThat` matcher correctly, this is just an alias that allows users to inline their assertion in the test which makes them read better IMHO leading to easier maintainability.

### DIFF
--- a/docs/5-stubbing-results.md
+++ b/docs/5-stubbing-results.md
@@ -182,6 +182,21 @@ yell('ARGHHHHHHH') // 'AYE'
 yell('ARG') // undefined
 ```
 
+###### Regular Expressions
+
+Using `contains` with regexps on string arguments is supported as of v1.11.0 and is also pretty straightforward:
+
+``` javascript
+var yell = td.function()
+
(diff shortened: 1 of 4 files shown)


searls:
Seems legit! This hadn't occurred to me, but it expands the inputs that `contains` can take without complicating the API any.

Will merge if you can just add documentation for the feature following `Strings` here: https://github.com/testdouble/testdouble.js/blob/master/docs/5-stubbing-results.md#strings

JaKXz:
@searls done :) hopefully what I've written makes sense to you, but feel free to tweak it to your preference on GH.

On a related side-note I'd recommend publishing a GitHub pages site for testdouble from the `/docs` folder, which you can do in Settings since y'all have them nicely organized.

searls:
Landed in 1.11.0. Thanks!

PR #179, opened 2017-01-17
chore: add editorconfig

JaKXz:
From one OSS maintainer to another, I find [EditorConfig](http://editorconfig.org) makes contributions a lot easier WRT whitespace and line endings etc etc. 

Up to you of course :)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,16 @@
+root = true
+
+[*]
+
+# Change these settings to your own preference
+indent_style = space
+indent_size = 2
+
+# We recommend you to keep these unchanged
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+
+[*.md]
+trim_trailing_whitespace = false


searls:
I'm not going to merge this at this time, for a few fair reasons and a few unfair ones:

## Filefile sprawl

I'm sensitive to the sprawl of `Filefile` and dotfiles that are expected to reside in every single project separately. That expectation (whether enforced by social norms or tools) subtly increases the cost of starting new modules and repos, and I prefer to keep that cost low, because myself (and most people) have too-strong-a bias for instead adding to an existing repo out of convenience. Moreover, when someone first looks at a repo, they tend to take stock of how many files exist to describe it, and top-level Filefiles/dotfiles can overwhelm people

## Personal vs Shared problems

`.editorconfig` solves the problem of "I work in multiple projects and they have different styles". If `.editorconfig` were a ubiquitous standard, supported by most people's editors  out of the box, then it might make sense for this to live in each repo. However, since only a small minority of developers work in an environment where `.editorconfig` is automatically loaded by their editors, it's not much different in practice from an IDE configuration file.  Put differently, `.editorconfig` may not be tool-specific, but if it's market share is 5% of a dozen different editors, then it's functionally equivalent to a proprietary IDE settings file, except for being bound by a social circle instead of by a specific tool.

One reason we don't commit `.project`, `.vscode`, `.idea`, `.vimrc` to every project is because of the sprawl problem mentioned above. Another reason is because by codifying the rule in a way that only affects a minority of users, you actually now have a DRY violation in that the rules exist in at least two places: a file somewhere and a separately-existing social norm. And those two things can fall out of sync with one another (as they already have in this PR in fact, because I configure `*.md` files to hard break at 80 characters but this file specifies they shouldn't be). 

`.editorconfig` absolutely solves an individual problem today, and I'd encourage you to use it if it makes you work faster. But it doesn't yet solve a shared problem (in the way package.json does), because our tooling doesn't ubiquitously support it. As a result, I'd prefer folks just set it and ignore it, as we would expect of IDE settings.

## Usefulness

Separately, I'm not convinced of the efficacy of such a file to begin with to actually make it easier to jump from project to project. In fact, even if widely adopted, `.editorconfig` could weaken community conventions over time.

To illustrate, if I had a `.facts-of-life` file and everyone had a tool that read the `.facts-of-life` file and adjusted my perception of reality to those facts, and the facts were:

```
water: wet
sky: blue
```

It would indeed be supported everywhere and tell people some true things about the project, but it wouldn't solve a problem most people have. (Except for folks in the winter in the midwest US, since sky is obviously `gray`.)  Additionally, some rebellious youth would probably set `water` to `dry` on their projects because they once heard that [DRY](https://en.wikipedia.org/wiki/Don't_repeat_yourself) is an important programming concept.

My point being that I have a strong bias for convention over configuration. In nearly every repo I encounter, I see 2 space indents and no-trailing-whitespace being the norm across the vast majority of projects I work in. I would much rather continue to push this as a strong social convention than to try to enforce it with a tool, as this PR would do.

The heart of the issue is that when you publicize that something is configurable, a certain percentage of people will go out of their way to configure it, sometimes only for the sensation of flouting societal defaults. So by having a file that declares these two basically ubiquitous rules, you'd be incidentally inviting people to tweak their *other* projects to suit their taste (say, 4-space indents). But that would run counter to the goal of reducing the cost of switching between projects, especially when you consider that `.editorconfig` tooling is not ubiquitous! 

To reiterate, popularizing a sparsely-used config file like this could actually have a stronger effect in weakening social norms than it does in increasing developer productivity.

## Conclusion

Ultimately, I wouldn't want to give others a convenient way to deviate from 2 space indents and trimming trailing whitespace, because I think we'd be better off if every open source JavaScript project adopted that style out of unconfigurable peer pressure. By not accepting this PR, I am encouraging you (normally implicitly, but to make this point, explicitly) to go and update your other JavaScript projects to match 2 space indents and to trim trailing whitespace. 

I think that's the more workable solution for each community in the long run (unless and until a `jsfmt` tool gains ubiquity like `gofmt` or Rust's formatter.

PR #190, opened 2017-02-28
Improve the error message when using Proxy incorrectly.

Schoonology:
Including a "did you mean" line. Shiny.


searls:
landed in 1.11.2

Issue #161, opened 2016-12-12
Better verification messages


searls:
#155 got me thinking about an impedance mismatch that exists in our messages.

Here's the inkling of an idea:

Argument matchers are a destructive operation when mapped over a series of calls. If I have calls `foo(1)`, `foo(2)`, `foo(2)`, and `foo('x')` and `td.verify(foo(td.matchers.isA(Number)), {times: 4})`, then the verification is going to fail.

Currently if this happens, this is the message you get:

```
Error: Unsatisfied verification on test double.

  Wanted:
    - called with `(isA(Number))` 4 times.

  But was actually called:
    - called with `(1)`.
    - called with `(2)`.
    - called with `(2)`.
    - called with `("x")`.
```

We were originally thinking we'd improve this message by batching identical calls, but that would result in batching the two calls to `foo(2)` and not the call to `foo(1)`, although it also satisfied the matchers. 

Therefore, I propose this satisfication-first format for verify messages:

```
Error: Unsatisfied verification on test double.

  Wanted:
    - called with `(isA(Number))` 4 times.

  3 calls that satisfied this verification:
    - called 1 time with `(1)`
    - called 2 times with `(2)`

  All calls of the test double, in order were:
    - called with `(1)`.
    - called with `(2)`.
    - called with `(2)`.
    - called with `("x")`.
```

Schoonology:
This looks amazing. :+1:

samjonester:
That is an elegant solution. It might even eliminate logical branching around how to present the "Actual" calls.

searls:
Fixed by #194

searls:
This somehow never got closed but it happened!

PR #194, opened 2017-03-06
Creates Better Verification Message

samjonester:
Adds a section to the unsatisfied verification message to describe
calls that match the verification. This section is displayed when the testdouble is
configured with a `times` that is greater than the number of matching
calls.

Example:
  Wanted:
    - called with `(isA(Number))` 4 times.

  3 calls that satisfied this verification:
    - called 1 time with `(1)`
    - called 2 times with `(2)`

  All calls of the test double, in order were:
    - called with `(1)`.
    - called with `(2)`.
    - called with `(2)`.
    - called with `("x")`.

This closes: Better verification messages #161
completed with @tinney


searls:
Looks good to me!

Issue #181, opened 2017-01-23
stub also verified warning when using ignoreExtraArgs


ackerdev:
When attempting to create reusable doubles, I found an oddity with the behavior of the "test double was both stubbed and verified" warning when using `ignoreExtraArgs`.

As a simple (contrived) example:
```javascript
const foo = td.function();

td.when(foo(), { ignoreExtraArgs: true }).thenReturn(bar);
td.verify(foo());
```

The intention behind this is to stub behavior of foo that consistently returns bar, regardless of arguments, while the verify call is to ensure that foo was called with 0 arguments. Because the stubbing is defined without arguments, the warning is displayed, even though `ignoreExtraArgs` should imply that these invocations are not 1-to-1 identical. It is possible to work around this with `td.matchers.anything()` in the stubbing, though not ideal.

My usecase for this is creating a reusable double to stand in for a somewhat complex third party interface that is passed in as an argument to my function; specifically, hapi's `reply` interface.

This is seems somewhat related to #180, a functionality limited by the stubbing api, though I think it's a distinct issue worth documenting.

searls:
This particular warning has been discussed in a lot of other threads if you want more on my thinking about this.

But I would suggest based on the other thread we were in yesterday that your use of the library is far enough outside the workflow I had in mind when designing it that you'll want to toggle the `ignoreWarnings` bit [here](https://github.com/testdouble/testdouble.js/blob/master/docs/C-configuration.md#tdconfig).

ackerdev:
Just to be clear, this issue isn't about the warning in general; I understand and agree with it. My point is that there is somewhat inconsistent behavior around the warning when using `ignoreExtraArgs`.

eg:
```
td.when(foo(), { ignoreExtraArgs: true }).thenReturn();
td.verify(foo()); // Warning!

td.when(bar(anything()), { ignoreExtraArgs: true }).thenReturn();
td.verify(bar()); // No warning

td.when(baz(), { ignoreExtraArgs: true }).thenReturn();
td.verify(baz(1)); // No warning
```

It seems the warning isn't taking into account the `ignoreExtraArgs` option. I'm not sure how it should behave even so, but just looked like it might be an unintentional inconsistency.

searls:
This is indeed inconsistent and I can't explain it.

samjonester:
@ackerdev Thanks for reporting the inconsistencies!

Here is my feeling about these cases:

``` js
// 1
[...]

PR #196, opened 2017-03-06
Improves warning when verifying a stub

samjonester:
After a verification is made, a warning can be produced to signal that
the verification call matches a stub on the same testdouble. This
behavior is improved to test the verification against the stubbing using
the stubs matchers and config options.

This closes: stub also verified warning when using ignoreExtraArgs #181
completed with @tinney


searls:
boom! awesome!

Issue #106, opened 2016-06-03
Support invoking callbacks with arbitrary timing


Schoonology:
Since all callback invocations occur synchronously, any code that assumes synchronous execution will break when exposed to the Zalgo of the real world.

It would be nice if `td` could be configured to invoke callbacks with different timings, either synchronously or not. I don't have a solid suggestion for that API, but would be happy to think on it if y'all agree this would be useful.

searls:
## Initial reaction

Agree with you that this is something worth considering. I'm, as usual, of two minds about this:

1. On one hand, td.js is designed to facilitate collaboration testing which isn't intended to provide rugged confidence for things like timing risks like this. I want to guard against adding features to td.js that would encourage testing patterns better left to a more integrated test.

2. On the other hand, it's entirely possible to test-drive these collaborators in such a way as to accidentally work synchronously, and since that's not how they'll ever be invoked in production, that's not good either.

3. On the third hand, I really like sync-by-default because I don't have to worry about silly `done` callbacking and so forth in my collaboration tests.

So where does that leave us? To be honest, I'm not really sure. 

## My first choice

My ideal, I think, would be a way to globally make all callbacks defer via some `setTimeout` or even `nextTick` as a way to just peek at the whole test suite every now and then that there aren't sync/async issues. However, that wouldn't really work, because:

1. most test frameworks have different semantics for sync tests from async tests, and each test would need to be updated for it
2. if there's a global option, some percentage of people will permanently toggle it on, despite the fact it's an unhealthy default (IMHO) for most uses and would be surprising to users visiting a project (as new tests might silently pass and the developer wouldn't know they need to code them as async)

## My second choice

The more practical option may be to add a `td.when` option, like you suggest, so that it's explicit and inline. I don't mind the [syntactic vinegar](http://david.heinemeierhansson.com/arc/2006_10.html) of making this hurt a bit, since I feel like it should be the exceptional case. The only problem is that most people will never see it or use it as a result, even though this is a risk that all callback-ey code faces.

I would propose:

``` js
td.when(schoon('hi'), {callbackTimeout: 5}).thenCallback(null, 'sup')
```

## My third choice

If this is a real problem and people are test-driving callback-APIs and bugs are introduced because td.js executes them synchronously, then I should give up my sync-test dream and just invoke the callbacks asynchronously to guard people against that risk by default. I really don't love this option but I'd do it if there's evidence that td.js is causing pain.

rmg:
[...]

PR #198, opened 2017-03-07
Adds config options for async callbacks.

samjonester:
To better represent the asynchronous nature of JavaScript development,
this commit adds two new configuration options for stubbing methods that
accept callbacks.

1. {delay: 10} - the delay option will delay invocation of the
callback provided to a test double by the provided number of
milliseconds
2. {defer: true} - the defer option will defer invocation of the
callback provided to a occur after the current callstack has cleared.

This closes: Support invoking callbacks with arbitrary timing #106


samjonester:
Depending on whether this API is acceptable, documentation will need to be added.

As an alternative API, `td.when().thenReturnAsync()` + `td.when().thenCallbackAsync()` might be more discoverable. They also might prevent confusion about when these new config options are valid.

test/src/callback-test.coffee
```diff
@@ -2,52 +2,87 @@ describe 'td.callback', ->
...
+        And -> @results == undefined
+        And -> @callbackInvoked == true
+
+      context 'EDGE CASE: thenCallback used but not satisfied', ->
+        Given -> td.when(@testDouble('/bar')).thenCallback('a-ha')
+        Given -> td.when(@testDouble('/bar')).thenReturn('o_O')
+        When -> @result = @testDouble('/bar')
+        Then -> @result == 'o_O'
+
+    context 'callback is asynchronous', ->
```
samjonester: I couldn't think of a cleaner way to represent the branching in this test. I'm up for suggestions.

src/store/stubbings.coffee
```diff
@@ -42,15 +42,24 @@ invokeCallbackFor = (stubbing, actualArgs) ->
   return unless _.some(stubbing.args, callback.isCallback)
   _.each stubbing.args, (expectedArg, i) ->
     return unless callback.isCallback(expectedArg)
-    callbackArgs = if expectedArg.args?
-      expectedArg.args
-    else if stubbing.config.plan == 'thenCallback'
-      stubbing.stubbedValues
-    else
-      []
+    args = callbackArgs stubbing, expectedArg
```
searls: if you wouldn't mind, all the single-line function invocations use `()`s in this project

test/src/callback-test.coffee
searls: Branching in this way is fine given that there's a shared `when` tying your hands at the top.

test/src/callback-test.coffee
```diff
@@ -2,52 +2,87 @@ describe 'td.callback', ->
...
+      When (done) -> @returnValue = @testDouble '/foo', (er, results) =>
+        @callbackInvoked = true
+        @er = er
+        @results = results
+        done()
```
searls: I don't think these tests do anything to ensure that the result is actually deferred, do they? What if you added a line at the end to make sure the results weren't set yet:

```coffee
     When (done) -> @returnValue = @testDouble '/foo', (er, results) =>
          @callbackInvoked = true
          @er = er
          @results = results
          done()
        @invokedSynchronously = @results?
    Invariant -> @invokedSynchronously == false
```

Maybe that'd do the trick?

samjonester: Seems like valuable addition. Definitely worth a try!

src/store/stubbings.coffee
samjonester: +1

PR #203, opened 2017-03-11
Support deep matchers

searls:
Resolves #142, #109, #160 

- [x] fix unit tests in #142 
- [x] patch stringify-object to be able to pretty-print deep matchers
- [x] fix circular structure bug (e4ea059)
- [x] fix entire build
- [x] refactor original PR
- [x] release stringify-object (or a fork)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,9 @@
     "prepublish": "yarn run compile"
   },
   "babel": {
-    "presets": ["env"]
+    "presets": [
+      "env"
+    ]
   },
   "browser": {
     "./lib/replace/module.js": "./lib/replace/module.browser.js",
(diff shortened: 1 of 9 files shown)


searls:
Figured out the reason for the breaking build is that when stringify-object dropped Node < 4 support, our browserify build will now contain a bunch of unsupported-in-some-browsers-like-phantomjs keywords like `const`.

I'm unsure how to tell `babelify` to transform a specific module dep, researching now. (cc/ @davemo @jasonkarns )

Reference:
https://github.com/yeoman/stringify-object/commit/76a5122bce1aea6b74fb13cc143cda6ce19d71cb

uhoh-itsmaciek:
omg thanks!

PR #206, opened 2017-03-14
Decaffeinate

searls:
- [x] convert src coffee to js
- [x] autofix standard js
- [x] pull in common-tags
- [x] manual remaining standardjs fixes
- [x] beat the ugly out of all these generated files.
- [x] do a pass of converting `let`s to `const`s 
- [x] do a pass converting `function` to `=>` where possible


searls:
Ugh — even though common-tags points to a babel-compiled `main`, it does _something_ that breaks on phantom@1.9.8, which makes me wary of including it, since I'm worried it'll increase the lowest supported browser. However since phantom@1.9.8 can't run on sierra it's not easy to test/debug what the root cause is.

For now I've removed common-tags but I'm a bit stuck as to how to proceed. @jasonkarns?

PR #209, opened 2017-03-17
Preserve function context when calling thenDo

connor4312:
This is required for some APIs where we might not have access to, but require (e.g. in fluent chains) the instance the function is being called on ahead of time.

You can be double-sure this works because I implemented it in both CoffeeScript and JavaScript, as #206 was merged while I was writing it 😉


searls:
Great catch, landed in 2.0.3

PR #211, opened 2017-03-18
Rework test build

searls:
There is a lot of work needed to whip the test build into shape, however we plan to do that (ES6, teenytest, etc).

- [x] get the tests built with one big browserify bundle instead of babel/coffee filter operations
- [x] upgrade our test deps (testem, mocha 3)
- [x] upgrade to browserify 14
- [x] upgrade to phantom 2 on travis (sick and tired of dealing with 1.9.8 quirks, compatibility check or no
- [x] remove the global-ish hacks and just use normal require where we can (this had been done because the tests had to run where `require` couldn't be used, but getting a browserify build solves this
- [x] audit all the tests that are currently flagged to run only when NODE_JS is true, as some of those should now be runnable


searls:
heads up @jacobthemyth, this actually worked! Took a ton of fiddling, but the build is now against the full browserified bundle

jacobthemyth:
Nice! I'm not sure how much I helped, but definitely happy to see it working :)

PR #215, opened 2017-03-21
Add failing test for resetting replaced modules.

Schoonology:
I think we discussed this case a long time ago, but this test fails when trying to `td.reset` a quibble-replaced module: the test double is still there.

### DIFF
--- a/test/src/replace/index-test.coffee
+++ b/test/src/replace/index-test.coffee
@@ -186,3 +186,7 @@ describe 'td.replace', ->
       describe 'and classes on objects on funcs', ->
         When -> td.when(@lights.brights.prototype.beBright(1)).thenReturn('yow')
         Then -> (new @car.lights.brights).beBright(1) == 'yow'
+
+    describe 'post-reset usage', ->
+      Given -> td.reset()
+      Then -> @car.honk.toString() == "throw 'honk'"


searls:
Ok, before we get started I think we need to disambiguate this, because this test case points out two issues:

1. This test's implied expectation is that `td.reset()` will immediately overwrite the fake things on `this.car`, but my expectation differs, because the `td.replace` mechanism means that the entire `this.car` is fake, and the _real_ car was never loaded, so any reference to `this.car` will continue being fake. I don't think the test _as written_ demonstrates a bug / deviation from my expectations.

2. However, when I tried adding `td.reset(); this.car = require(''../../fixtures/car')`, I _was surprised_, because I would have expected `td.reset()` (which calls `quibble.reset()` to unwind the `Module._load` intercepts and re-enable the cache. 

I suspect you ran into this as a result of test pollution? (i.e. two tests requiring the same module and one getting fakes placed for the purpose of the other?) That seems like what you'd experience if everything above is accurate.

So as a result this does seem like a bug in either td.js or quibble. My expectation is resetting should make it safe to immediately re-require the dependency and get the real, require-cacheable copy of the dependency until/unless it's re-quibbled.

Schoonology:
This is an older test, but it was in a scenario where I had one suite `td.replace`-ing another internal module, and another suite attempting to test that module. The latter suite got the double, and "hilarity" ensued.

The next time I get a chance to look at this type of situation, I'm hoping a pattern around how to guarantee that _re-require_ aspect happens appropriately will emerge. I'm not convinced I had that set up correctly.

searls:
Fixed this in quibble. Was annoying. The end.

searls:
Landed in 2.0.5

Issue #48, opened 2015-11-14
td.explain should work on objects that contain test doubles


searls:
If td.explain is passed an object, it should loop over the functions and, if any test double functions are found, should report some rollup stats (e.g. "my name is this, here are the test doubles, they have X many calls", etc)

Thanks @johnkpaul for the idea!

robwold:
Hi, 
I've been using testdouble all of 1 week, but I ran into this issue on day 1. (I'm enjoying using it though, so thanks for your hard work.) Having checked the source code for `explain` I'd be happy to submit a PR based on the approach quoted above:

- test the `typeof` argument passed to `explain`; 
- if it's an object, check if each of its values is a test double function (or some nested object with a test double function in there somewhere);
- produce a summary report by applying the existing method for `explain` to each test double in turn.

However, in the issues that reference this one, the discussion seems to take place at the level of the interaction between `explain` and "the store", which I wouldn't touch with this approach. So as I've not ventured far into the codebase, and this issue has been open a while, I thought I'd ask before I spent the time writing the code: is there a reason you'd prefer to avoid the rock-dumb approach suggested here, or is there some reason why it wouldn't work that I'm missing?

searls:
Hey @robwold thank you for the offer! To ensure your time is well spent, how about you send a PR with a quick and dirty proof of concept and then we can discuss that before figuring out how to best incorporate and test it (re: other design concerns)

PR #217, opened 2017-03-23
Add example failing test for td.explain(object).

Schoonology:
This is just another random thought, but I found myself instinctively trying the following while test-driving a feature this week:

```
var bag = td.object(['some_method'])
...
td.explain(bag)
```

While it was immediately obvious both that we don't support this behaviour now and that what I wanted was `td.explain(bag.some_method)`, it left me wondering how desirable this behaviour would be. I know it would require substantial changes to the way `explain` and `store` interact. (If it was easier, this PR would include the working feature!)

### DIFF
--- a/test/src/explain-test.coffee
+++ b/test/src/explain-test.coffee
@@ -95,3 +95,17 @@ describe '.explain', ->
       callCount: 0
       description: "This is not a test double."
       isTestDouble: false
+
+  context 'a bag of doubles', ->
+    Given -> @testDouble = td.object(['example'])
+    Then ->
+      console.log @result
+      expect(@result).to.deep.eq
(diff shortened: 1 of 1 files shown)


searls:
We've had this issue logged for a long time. https://github.com/testdouble/testdouble.js/issues/48 

I think it's possible without _too_ much rework. We can be a little deft about it if we just slam these objects/constructors into the store with added references of the test doubles that dangle off of them, and then branch accordingly in `td.explain`. 

It'd be hacky to start storing heterogeneous data into the store so it might make sense to instantiate a second store and read from that one when `td.explain` is a miss on whatever's passed to it.

searls:
Closing as stale

PR #227, opened 2017-04-08
Rollup

searls:
**Not ready to merge**

- [x] smell test seems to work in browsers with `yarn run compile:browser; cat dist/testdouble.js | pbcopy`
- [x] actually make build work
- [x] make unit tests work
- [x] make sure node cjs build is tested via example
- [ ] make sure node es build is tested via example
- [x] make integration tests work
- [ ] make example tests work


searls:
Current issue:

1. Because `process.cwd` was being overwritten with `/` whenever I added `globals()` or `builtins()` which I really don't want for the node.js cjs build (via `es-process`), I decided to branch on whether to include them at all based on whether it was a browser build

2. This actually got me to a passing unit test & browser test build, and only blew up once I got to my webpack example build. 

3. It does causes a bunch of warnings, as you might expect, so I guess I'm doing it wrong?
screen shot 2017-04-08 at 2 47 17 pm

4. It causes a problem with the webpack build because `fs` isn't defined, since webpack is loading the `cjs` build and the runtime lacks `fs`. At runtime it won't hit it, but rollup pulls those require statements to the very top of the build so webpack blows up.
screen shot 2017-04-08 at 2 50 52 pm


To reproduce and help out: `yarn run test:ci`

Taking a break b/c this is a mind-numbing work

searls:
We may still adopt rollup but the build has changed enough that this won't help much

PR #237, opened 2017-05-04
Fix #236; Add another overload and its test

### LINKED ISSUE #236: TypeScript: Better Generic Support for td.object()
Given the following... 

```
import * as td from "testdouble";

class FooBar {
  private _date: Date;
  constructor() {
    this._date = new Date();
  }
  public dateString() {
    return this._date.toString();
  }
}

let fooBarFake = td.object<FooBar>("FooBar" as any);
```

The above code will set `fooBarFake` to FooBar object as expected. TypeScript and `fooBarFake` behaves as I expect too. However, without `as any`, TypeScript will throw the following error...

`'"FooBar"' is not assignable to parameter of 'FooBar'`  

Take *notice* of the quote marks.  Without the generic parameter (`<FooBar>` ), TypeScript `fooBarFake` thinks it of `String` type.

I believe the parameter should not need to be cast as `any` type for the generic parameter to be accepted. 

```
import * as td from "testdouble";

// define FooBar class here

let fooBarFake = td.object<FooBar>("FooBar");
```

*As side note, I generally set TypeScript's `noImplicitAny` setting is to true.*

sgtoj:
Ran `test:typescript` npm script before committing.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -20,6 +20,9 @@ export function object<T>(constructor: { new (...args: any[]): T }): DoubledObje
 // When passed array of props
 export function object<Key extends string>(props: Key[]): DoubledObjectWithKey<Key>;
 
+// When passed class or constuctor function name as string value
+export function object<T>(object: string): DoubledObject<T>;
+
 // When passed general object
 export function object<T>(object: T): DoubledObject<T>;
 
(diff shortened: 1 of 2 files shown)


searls:
Once again merging without understanding ✨

Issue #238, opened 2017-05-04
TypeScript: Improve Definition


sgtoj:
List of improvements I believe is needed for the TypeScript definition. I have already started working on this list. Love to hear feedback if you have any.

- [ ] Organize / Restructure  
- [ ] Add JSDoc for all functions
- [ ] Define `td.func()`
- [ ] Add generic types to `td.function()` and `td.func()`
- [ ] Define `td.constructor()`
- [ ] Add to `./test/src/typescript/test.ts` to reflect any additions.

searls:
Any improvements are appreciated. Hey @duluca, you'd suggested helping out with our typescript story, want to look at this?

sgtoj:
Updated OP with the link to the fork and branch.

sgtoj:
PR #239

duluca:
Will review. @sgtoj please lmk if there's anything specific I can help with. I'll leave any comments re: code on the pr

sgtoj:
Disclaimer: I am not the best when it comes to communication so there may be the grammar or typos. I did a quick proof reading before committing. 

The jsdoc comments probably could use some addition or improvement from someone who has more experience with the framework and/or understanding behind the abstractions. Otherwise, it should be ready.

duluca:
Sounds good. Will give it a shake tonight.

albertogasparin:
Sorry to jump in, but with published typings I'm getting an error in VSCode (Typescript 2.3):
```js
const myMock = td.object(['run', 'search']);
td.when(myMock.run()).thenCallback(null);
// Error: Property 'run' does not exist on type 'DoubledObjectWithKey<"run" | "search">'.
```

Is it something this PR is going to solve? Or are you guys not getting this?

searls:
@albertogasparin I suggest you attempt checking out the branch in #239 to find out for yourself

albertogasparin:
Yep, was doing it. In fact, I'm still getting the same error even with the PR typings. Unfortunately I don't have much experience with Typescript, so I'm unable propose a fix

sgtoj:
@albertogasparin I will look in to this issue. Hopefully, I will find a fix, add a test, and submit a new PR.

sgtoj:
I fixed the issue and added a new TS test in #239. @albertogasparin @searls 

![Example](http://i.imgur.com/6MctLeg.png)

albertogasparin:
@sgtoj Worked like a charm, thanks! 
[...]

PR #239, opened 2017-05-04
Improve TS Support

sgtoj:
See #238 for more information.


duluca:
@sgtoj I've started the review, this may be more productive if I sent you a PR to fix some of the grammar stuff, which will also give me a chance run the code and look at how things are fitting together. I'm getting ready for a conference talk for tomorrow. So expect to hear from me again sometime Saturday!

duluca:
@sgtoj @searls not sure if my testing methodology is correct here, but i'm running into an error. I tried to execute `test/src/typescript/test.ts` using `node` after running `tsc` on it to transpile it to a `.js` file. 

Here's the error message that gets thrown when `const bear = td.object<Dog>("Bear");` executes
```

:43:4)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.runMain (module.js:604:10)
```
I then attempted to dig into the `object<T>` declaration and lost the trail with `export type DoubledObject<Subject> = Subject;` I'm unable to find where Subject is defined.

I'm still getting oriented with the code base. Some help and direction would be appreciated.

sgtoj:
@duluca `Subject` is defined as the generic parameter. So... 

```js
(DoubledObject<Subject> = Subject) === (DoubledObject<T> = T)
```

sgtoj:
@duluca I would prefer to use `T`, though.

duluca:
@sgtoj Ah I see, naming generics other than T is usually then when you limit what types can be used by the generic. If you can use any object, T or TObject makes sense. If say, the object must implement an interface called IDocument, then calling it TDocument helps the user understand what's expected. So here I'd recommend going with T or TSubject. However having an ISubject defined would be helpful, if you expect the object to have specific shape. But I also don't get what `DoubledObject<T> = T` accomplishes :)

duluca:
@sgtoj So here `export type DoubledObjectWithKey<T extends string> = { [K in T] };` it is fair, and actually quite helpful to call it TString.

sgtoj:
`DoubledObject<T> = T` was `DoubledObject<Subject> = Subject` before I changed it to be consistent with the rest of the definition. I am not sure what the original author's motivation behind the naming of the types defined at the top.  

Although I normally see single-letter -- T, K, V, etc -- type parameter names, I am open to the idea. 

I just looked where the letters originated. Apparently, this convention is adopted from [Java](http://stackoverflow.com/questions/2900881/generic-type-parameter-naming-convention-for-java-with-multiple-chars) and C#.

searls:
I'm not a Typescript user, but I agree that generic types should use single-letter placeholders like `T` and `K`,`V` as those are overwhelmingly common conventions and point out to users that they are customizable.

The real tricky bit that I would want to accomplish from a type awareness for td.js is for any created test double to have its type inferred or specified such that a compilation failure is triggered until the type is created. Same goes with creating a stubbing which then fails compilation and requires the method be written.

To express this with code, what I want is for it to support a workflow like I would have in Java + Mockito. Here's how that would normally flow

``` java
public void testInvoices() {
  // 1. Write the fake you want
  LoadsPayments loadsPayments = mock(LoadsPayments.class) 
  // 2. get compiler feedback that you need to go define LoadsPayments (e.g. hit ctrl-1 to quick-fix create the type

  // 3. Write a stubbing of the method you want the fake to define
  Payment payment = new Payment()
  when(loadsPayments.load(new Date())).thenReturn([payment])
  // 4. get compiler feedback that `LoadsPayments` needs an instance method `load` with parameters `(Date)`  and return type `Array<Payment>`, prompting me to go create that in a way that my tools can auto-generate for me

  // … rest of test goes here



}
```

This sort of paint-by-numbers approach where writing a single isolated test informs the tooling with what it needs to create all the files/classes for you is a fantastic productivity win, and one that encourages lots of small custom types.

How possible is this going to be with TypeScript?

duluca:
Yeah I saw, that was a great change. And I have more C# experience than I care to admit. Given the same guy invented C# and TypeScript, I think it's fair to carry over the conventions. And naming generics start with a capital T (i.e. TSubject or TString) is similar to how you name interfaces with a capital I (i.e. ISubject). 

K and V are stand-ins for Key and Value. Personally I prefer TKey and TValue.

duluca:
@searls I know that we can do that easily if we rewrite portions of the app in TypeScript, not sure if it can be achieved through type definitions alone. But then the library can be compiled down to JavaScript and continue to used without any notion of TypeScript by consumers, if they choose to.

searls:
I am entirely okay with this

sgtoj:
I am also interested on that project too. May do so work towards it weekend.

duluca:
@sgtoj @searls I've noticed that `yarn test:typescript` merely transpiles the file, doesn't attempt to run it. Thus not an adequate test harness.  
So,
1. We need to implement test script, which will execute the code
2. Fix the test code so `ln 58: td.replace({}, "prop");` doesn't error out
3. Only then can this PR be considered as verified by the CI process

@sgtoj let me know if you need help

sgtoj:
@duluca Sorry... I will review and fix tonight or this weekend. So far, I have already these TS definitions for 2 different work projects; with 2 more in the works. I haven't found any issues yet.

duluca:
In `package.json` the property `"test:typescript"` should look like: `"tsc -p ./test/src/typescript && node ./test/src/typescript/test.js && rm ./test/src/typescript/test.js"`

duluca:
@sgtoj I fixed up test.ts in #259, also updated package.json to actually execute the code. So once the PR can be merged, it'd be great to update this PR from master, so there's a valid CI run against this change.

TODO:
[ ] Update from master once #259 is merged
[ ] Update this PR and make sure CI checks pass

searls:
Ping that #260 is merged -- should this be revisited?

duluca:
@sgtoj do you need help with this?

sgtoj:
I have merged from `upstream master`, fixed one conflict, and fixed one problem pointed out by the test.

sgtoj:
That makes sense. I was very confused too. I will fix.

searls:
@duluca please advise if you would like to merge and what version segment is appropriate to bump

duluca:
@searls It makes sense to me as a patch increment.

PR #242, opened 2017-05-06
Avoid calling super() when extending ES classes

### LINKED ISSUE #241: Figure out babel+instanceof+constructor problems with class extension
Moving #157 here now that we know what's going on.

Three things at once going on here: when you replace a class, td.js will extend that class and give you a fake one.

1. we extend the class so the fake will pass `instanceof` checks in the subject. This was raised by several issues in td 1.x and was the main motivation for the change
2. because of ES rules, if that class has a custom constructor, subclass constructors _must_ call `super`, which sucks because a fake thing shouldn't be invoking a real constructor
3. because of how babel transpiles, faking a real class will extend it from a not-real (transpiled babel) class, which calls the super constructor but without the `new` keyword, which raises an error under v8 versions that support the `class` keyword 

Dupes: #240

searls:
Got the idea from: http://stackoverflow.com/questions/31067368/javascript-es6-class-extend-without-super

Tested it in this runkit:

https://runkit.com/searls/avoiding-calling-super-in-an-extended-class

Resolves #241

### DIFF
--- a/src/constructor.js
+++ b/src/constructor.js
@@ -13,8 +13,9 @@ var fakeConstructorFromType = (type) => {
 
   return _.tap(class TestDoubleConstructor extends type {
     constructor () {
-      super(...arguments)
+//      super(...arguments)
       fauxConstructor(...arguments)
+      return Object.create(new.target.prototype, {})
     }
   }, (fakeType) => {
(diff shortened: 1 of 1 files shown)


searls:
Resolved by #254

PR #247, opened 2017-05-15
Add isClass ES6 class check for td.object

razor-x:
@searls This is an attempt to address the regression in https://github.com/testdouble/testdouble.js/issues/241#issuecomment-299611507. 

This will not work if the code you are trying to test is first converted with babel to a target that does not support `class` as then you would be passing a `function` to `td.object`. I'm not too concerned with that as right now it's broken if you are using `class` where it should be supported.

Maybe there is a better solution, but I would like to see the regression patched.


razor-x:
Also, I'm not sure how to add tests for this since they are in coffeescript and this is specifically targeting an ES6 feature that conflicts with `class` from coffeescript.

searls:
Close for reason mentioned in #241

PR #259, opened 2017-06-14
Enabling tsc testing pipeline

duluca:
- Changed package.json command to build `test.ts` into `test.js`, use
node to execute the file to verify correct function, then clean up the
file
- Updated gitignore to ignore 
`test/src/typescript/test.js` to prevent accidental checkin
- Updated `test.ts` to fix broken or invalid test double code. It seems
that `.constructor()` now requires the usage of `.prototype` in front
of function names, not sure if this is a recently introduced bug or not.


duluca:
@searls The CI build seems to be failing due to an unrelated error
```
browserify src/index.js --standalone td --outfile $npm_package_config_build_file -p headerify -t babelify 
events.js:163
      throw er; // Unhandled 'error' event
      ^
Error: ENOENT: no such file or directory, open 'dist/testdouble.js.tmp-browserify-13020562542777103232'
```
Any idea why?

searls:
Why did you remove dist/testdouble.js? That's intentionally there for browser users, as it's what they should be depending on (whether thery're curling it, or installing and then linking to it)

searls:
Two other Q's:

* Does the VS Code specific stuff need to be in the repo? I generally ignore any editor-specific config files, as everyone uses different tools
* Why ignore a file in test/**? Is it because the file is generated? If so, I'd have it written to some path under `generated/` which is already used by other tests and already ignored. Ignoring a path under `test/` suggests to me that there is a source listing that's being left out for some reason

duluca:
@searls I was presuming any `dist` assets would be generated by the CI pipeline and it was popping in as a file change. But I'm beginning to gather, we're suppose to check-in if `dist/testdouble.js` gets updated?

searls:
`dist` is only checked in as part of `npm version {major|minor|patch}` runs, so that at any given point, `master`'s copy of `dist/testdouble.js` points to the most recent release version. If you run `yarn test:ci` locally and it updates the `dist/*` directory, just `git checkout dist` and be sure not to commit the changes, as we only want stable versions committed there on release tag refs

duluca:
- will remove `.vscode` I had pushed it because it contained a helpful debug config, but cmd line works better 
- will undo my checkin of `dist/testdouble.js`
- will utilize `generated` instead and clean up .gitignore

duluca:
Closing in favor of #260 -- since this PR touched `dist/testdouble.js` and I wasn't able to reconcile the history properly

PR #260, opened 2017-06-14
Enable tsc in ci pipeline take 2

duluca:
- Changed package.json command to build `test.ts` into `test.js`, use
node to execute the file to verify correct function, then clean up the
file (unfortunately it is not possible to use the `generated` folder to
do this, otherwise dev experience of `test.ts` breaks given the
relative location of `td`)
- Updated gitignore to ignore `.vscode` (I didn’t add
`test/src/typescript/test.js` since if tests are successfully passing
this file shouldn’t exist, but could be added as a defensive measure)
- Updated `test.ts` to fix broken or invalid test double code. It seems
that `.constructor()` now requires the usage of `.prototype` in front
of function names, not sure if this is a recently introduced bug or not.
- Added mocha and chai typings to make it possible to add actual unit
tests to `test.ts` — for now implemented a simple assert


duluca:
@searls now seem to be a different set of failures. Looked into it for a while but no dice. 
1. Node 4 build seems to fail due to 'no proxy support' -- given my local experiments, it seems that test double is not compatible with Node 4 (at least in the context of TypeScript compiling a class down to ES 5)
```
var Bear = (function () {
    function Bear() {}
    Bear.prototype.sleep = function () {}
    return Bear;
}());

var bear = td.object("Bear");
```
```
​​Error: testdouble.js - td.object - The current runtime does not have Proxy support, which is what​​
​​testdouble.js depends on when a string name is passed to `td.object()`.​​
```
2. Not sure why the Node 6 builds fails due to a seeming timeout issue, but Node 7 build passes

searls:
FWIW, I am able to repro the Node 4 failure locally. Disregard the Node 6 failure, I have a flaky unit test that's somehow triggering a timeout.

searls:
Fixed and merged (I made a couple commits)

duluca:
@searls awesome thanks

Issue #263, opened 2017-06-17
The Big Epic Rewrite Issue


searls:
Starting with the setup work in #255, I've been undergoing an in-place rewrite of testdouble.js that essentially dogfoods both the library & [Discovery Testing](https://github.com/testdouble/contributing-tests/wiki/Discovery-Testing) by replacing the library bit-by-bit.

This has been incredibly challenging for a variety of reasons, among them: keeping oneself unanchored by existing implementation when trying to think of a better design, coming up with useful names for (relative to most JS libs) very meta concepts, and resisting the urge to fix problems and incongruities in the existing implementation as I go. If this rewrite also substantially changes the public-facing behavior of the library, it's going to be like tuning two knobs on a radio at once, making everything more difficult. 

Therefore, this thread will maintain 3 to-do lists for tracking progress of the rewrite. First: the actual progress so far incorporating the new implmentation. Second: the future work to undertake to "fix" what I identify as broken with the library after we call the rewrite "done". Third: internal improvements that should be tracked and made as part of the rewrite.

## Progress so far.

testdouble.js has a pretty straightforward public-API:

- [ ] test double creation
  - [x] td.function (alias td.func)
  - [ ] td.object
  - [ ] td.constructor
- [ ] td.replace
  - [ ] CommonJS module replacement (Node.js only)
  - [ ] named property replacement
- [ ] td.when for stubbing
- [ ] td.verify for verifying calls
- [ ] matchers for both stubbing and verifying
  - [ ] the same built-in matchers we have today
  - [ ] the same matcher-creator function we provide today
  - [ ] the special matchers like `captor` and `callback` which enable other behavior
- [ ] td.reset for resetting test double state and undoing dependency replacement
- [ ] td.explain for providing metadata about a fake thing
- [ ] td.config for global config options (log level, which promise constructor to use, etc)
- [ ] Internal bits
  - [x] A logging facility
  - [x] An in-memory store for calls (`CallLog`)
  - [x] An in-memory store for stubbings (`StubbingRegister`)

*being labeled complete here does not mean they are wired up in the deployed library, yet. They'll sit in the repo as inert code until they can go end-to-end. For instance, `td.func()` can't be replaced until its calls go in the same call log that `td.verify()` is looking at, otherwise the library wouldn't work. It truly is a big epic rewrite as a result, since almost everything depends on everything else because of how stateful a library this is.

## Room to improve

[...]

Issue #224, opened 2017-03-28
Consider using _.isObjectLike rather than _.isPlainObject within td.object


traviskaufman:
Given this code:

```js
const td = require('testdouble');

class MyClass {
  doSomething() {}
}

function callDoSomething(myClass) {
  myClass.doSomething();
}

const mock = td.object(new MyClass());
callDoSomething(mock);
td.verify(mock.doSomething());
```

When run on a browser that doesn't support proxies - such as IE11 - this code will throw an error saying: 

```
Error: Error: testdouble.js - td.object - The current runtime does not have Proxy support, which is what testdouble.js depends on when a string name is passed to `td.object()`.
```

I believe this to be a false positive, with the offending code being https://github.com/testdouble/testdouble.js/blob/c54b74187431a991817816dde614e252a3ae1a13/src/object.js#L16

This code uses `_.isPlainObject`, which returns false for instances of objects whose `[[Prototype]]` is not `Object.prototype`. This is problematic for instances of classes, who's `[[Prototype]]` is the constructor. 

I'm wondering if you'd be willing to switch to [_.isObjectLike](https://lodash.com/docs/4.16.4#isObjectLike) instead of `isPlainObject`? This will return `false` for functions / constructors, but `true` for instances.

I'm able to work around the `_.isPlainObject` limitation by creating a fake constructor using `td.constructor`, and then instantiating that instead, but it is a bit more boilerplate. 

Thanks!

searls:
I think you are probably right and that `_.isObjectLike` is closer to what we desire here.

However, I'd like to understand why `td.constructor()` isn't a better fit for your needs in this case, given that your fake is an instantiated Something.

searls:
I dunno. I guess I could go either way after writing this out:

```js
const mock = td.constructor(MyClass);
callDoSomething(new mock());
td.verify(mock.prototype.doSomething());
```

searls:
Hmm, there are really three affected areas by this distinction and I think it warrants some thinking before we change anything.

- [ ] [td.matcher.contains](https://github.com/testdouble/testdouble.js/blob/master/src/matchers/builtin/contains.js#L12) - would have to put regex above and then ask ourselves if there are cases this could screw up
- [ ] [td.replace](https://github.com/testdouble/testdouble.js/blob/master/src/replace/imitate.js#L12), but this seems like it'd be safe since the only other path would be an error
[...]

PR #264, opened 2017-06-18
Allow stubbing of custom instances & Object.create()

### LINKED ISSUE #257: td.replace throws with objects created with Object.create
First off, I'm a big fan of the project, so thank you!

Second off, I'm trying to mock a module using `td.replace` that exports an object created with `Object.create`, but `td.replace` throws. I haven't dug into testdouble's code yet, but my gut says it has something to do with enumerability (because the function properties of the object are non-enumerable; maybe related to #119?). Is this expected behavior?

Reproduce with:

``` js
// index.js
const td = require('testdouble')

try {
  const obj = td.replace('./normal-object')

  console.log('Normal object')
  console.log(obj)
  console.log(obj.foo)
  console.log(obj.bar)
} catch (e) {
  console.log('Failed to replace normal object')
  console.error(e)
}

console.log()

try {
  const obj = td.replace('./non-enumerable-object')

  console.log('Object with non-enumerable properties')
  console.log(obj)
  console.log(obj.foo)
  console.log(obj.bar)
} catch (e) {
  console.error('Failed to replace object with non-enumerable properties')
  console.error(e)
}

console.log()

try {
  const obj = td.replace('./object-from-create')

  console.log('Object from Object.create')
  console.log(obj)
  console.log(obj.foo)
  console.log(obj.bar)
} catch (e) {
  console.error('Failed to replace object from Object.create')
  console.error(e)
}
```

``` js
// normal-object.js
module.exports = {
  foo: () => 'foo',
  bar: () => 'bar'
}

```

``` js
// non-enumerable-object.js
module.exports = Object.defineProperties({}, {
  foo: {
    enumerable: false,
    value: () => 'foo'
  },

  bar: {
    enumerable: false,
    value: () => 'bar'
  }
})

```

``` js
// object-from-create.js
module.exports = Object.create({
  foo: () => 'foo',
  bar: () => 'bar'
})
```

Output

``` shell
$ node index.js
Normal object
{ foo: { [Function: testDouble] toString: [Function] },
  bar: { [Function: testDouble] toString: [Function] },
  toString: [Function] }

Object with non-enumerable properties
{ toString: [Function] }
[Function: value]
[Function: value]

Failed to replace object from Object.create
Error: Error: testdouble.js - td.replace - "./object-from-create" property was found, but test double only knows how to replace functions, constructors, & objects containing functions (its value was "[object Object]").
    at Object.error (/Users/mc/sandbox/td-object-create/node_modules/testdouble/lib/log.js:23:13)
    at exports.default (/Users/mc/sandbox/td-object-create/node_modules/testdouble/lib/replace/imitate.js:45:26)
    at exports.default (/Users/mc/sandbox/td-object-create/node_modules/testdouble/lib/replace/module.js:12:41)
    at exports.default [as replace] (/Users/mc/sandbox/td-object-create/node_modules/testdouble/lib/replace/index.js:9:29)
    at Object.<anonymous> (/Users/mc/sandbox/td-object-create/index.js:34:18)
    at Module._compile (module.js:570:32)
    at Object.Module._extensions..js (module.js:579:10)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
```

searls:
* Extracts from #263 rewrite common mechanisms for identifying properties, copying them to target objects, and filtering them down to the list of functions on the fake. This will iron over a few inconsistencies between `td.func`, `td.object`, and `td.constructor`
* The new property-identifier will look at a target thing's enumerated & non-enumerated properties, inherited and non-inherited. It'll leave off any cruft inherited from `Object.prototype` and `Function.prototype`, however
* Any custom instance (e.g. `Object.create({foo: 1})`) will now be acceptable by `td.object`. Previous restrictions against this by me were specious / not defensible IMO

Fixes #257, #224


searls:
Landed in v3.1.0

PR #268, opened 2017-06-23
imitate / replace things deeply (instead of shallow)

### LINKED ISSUE #262: Single-layer property replacement is inappropriate for ES modules
RIght now, `td.replace`'d CJS modules enjoy a single-layer of additional properties (say, a function with other static functions as properties) being faked along with the top-level export.

This is incongruent with ES modules, because that first layer of replacements typically is just enough to cover the `default` property on the exported object (which is how transpilers translate those imports to CJS), which means the same module will be faked differently if all a person does is convert it a CJS module to an ES module.

Going forward I think we need to do one of two things:

1. Pick a depth and then make it feel consistent for both module types (e.g. perhaps we keep it as is, but add an extra layer of function fakes and property cloning for any export named `default`
2. Recursively fake whatever we find on a module, and do whatever work we need to break out of cyclic structures

Option 2 is probably more appropriate for simplicity & consistency sake, but it also runs the risk of encouraging folks to return deeply-nested mocks, which is separately a little worrying. That said, even that would be better than the current state: accidentally returning people partial mocks that have some properties that are still real (or missing)

searls:
Build on #257 by creating a single unit responsible for all "imitation" of real things, be they plain functions, objects, or constructors.

This does a few neat things for the library in terms of enhancement:
* It seems to fix the `instanceof`/inheritance issues that forced the 3.0 major bump. If this turns out to be true, it means the config property can be eradicated and we can put that whole mess behind us
* It will perform an arbitrarily deep imitation of whatever you pass it, cloning most things and replacing functions with test doubles. This will make it more suitable for ES modules
* It will remove the inconsistencies between `td.func`/`td.object`/`td.constructor`, in fact, most of the implementation of each of those can be thrown out
* Work fine when passed things containing cyclic references 

This is an incremental improvement that shook out of the Big Epic Rewrite in #263

## Concerns

If you are in the habit of calling `td.replace` on massive modules (with hundreds or thousands of cumulative properties), this is going to be _realllllllly_ slow compared to what you're used to, at potentially no added benefit. Trade-offs here:
1. The new implementation could surely stand to be optimized, since it's new,
2. You really shouldn't be replacing massive dependencies in a unit test, because (a) you shouldn't be writing massive dependencies, and (b) you shouldn't be mocking what you don't own (QED)
3. I'm open to the idea that a `maxImitationDepth` config setting may be appropriate (note that the current implementation successfully breaks cycles, but will otherwise recurse to arbitrary depth)

## Todo

I made a bit of a mess in the process of getting this over the hump and the build passing. Some to-dos remain:

- [x] eliminate src/share/filter-functions if it's not used
- [x] move the src/share/* functions under src/imitate
- [x] audit wrap/lodash and remove any unused exports
- [x]  validate that extendWhenReplacingConstructors is no longer necessary by getting an example test that uses `instanceof`, then remove any remnants of the option 
- [x] eliminate the prototype-aware bits of `copyProps` if imitate has successfully obviated it
- [x] rename `test/` to `/regression`
- [x] move `unit` to `test/unit` and fix all the imports
- [x] create `test/safe` and put the functional imitation test there
- [x] drive out a new unit test of the imitate function to iron out the varying levels of abstraction present
- [x] add a new regression test that covers some of the deep imitation / replacement characteristics

Fixes #262


src/constructor.js
```diff
@@ -1,54 +1,14 @@
 import _ from './util/lodash-wrap'
 import tdFunction from './function'
-import config from './config'
-import copyProps from './share/copy-props'
-import gatherProps from './share/gather-props'
-import filterFunctions from './share/filter-functions'
+import imitate from './imitate'
 
 export default (typeOrNames) =>
```
jasonkarns: I'd suggest naming the function so you'll get more useful stack.

src/function.js
```diff
@@ -2,24 +2,15 @@ import _ from './util/lodash-wrap'
 import calls from './store/calls'
 import store from './store'
 import stubbings from './store/stubbings'
-import gatherProps from './share/gather-props'
-import copyProps from './share/copy-props'
-import filterFunctions from './share/filter-functions'
+import imitate from './imitate'
 
 export default (nameOrFunc, __optionalName) =>
```
jasonkarns: same here (name the exported func)

src/function/index.js
```diff
@@ -1,14 +1,12 @@
 import _ from '../wrap/lodash'
 
 import create from './create'
-import imitate from './imitate'
+import imitate from '../imitate'
 import remember from './remember'
 
 export default (nameOrFunc) => {
```
jasonkarns: samesies

src/imitate/copy-props.js
```diff
@@ -0,0 +1,31 @@
+import _ from '../wrap/lodash'
+
+export default (target, props, visitor) => {
```
jasonkarns: more anon default exports...

src/imitate/create-imitation.js
```diff
@@ -0,0 +1,17 @@
+import _ from '../wrap/lodash'
+
+import tdFunction from '../function'
+
+export default (original, names) => {
```
jasonkarns: same

src/imitate/initialize-names.js
```diff
@@ -0,0 +1,16 @@
+import _ from '../wrap/lodash'
+
+export default (original, names) => {
```
jasonkarns: same

src/imitate/overwrite-children.js
```diff
@@ -0,0 +1,33 @@
+import _ from '../wrap/lodash'
+
+import gatherProps from './gather-props'
+import copyProps from './copy-props'
+
+export default (original, target, overwriteChild) => {
```
jasonkarns: same

src/replace/module.js
```diff
@@ -7,11 +9,16 @@ quibble.ignoreCallsFromThisFile()
 export default function (path, stub) {
```
jasonkarns: same

searls:
lol @jasonkarns why did you leave the same comment a million times.

FWIW, I tend not to name these functions since the stack trace gives me the file name & I tend to rename really aggressively, and naming the function adds yet another thing I have to change (in addition to file name, test file name, the test's import, and any production imports). Because I tend to create a lot of small things and change them pretty frequently, I tend not to value naming functions like this until well after the dust has settled on the design

jasonkarns:
Was just tracking all the spots so nobody else had to search :)



> since the stack trace gives me the file name

Didn't realize babel did this now (always?). Hoping it works for the browserified version.  🤷‍♂️

searls:
Re: babel, it depends on how you compile it. We compile via a filter operation into `lib` (meaning no concatenation occurs for node users) and we separately babelify for browser users. The former gives the pretty stack traces. The latter, well, I'll welcome a fixup PR some time after this big epic rewrite to sprinkle in all the right names

searls:
Landed in 3.2.0

PR #274, opened 2017-06-30
Don't blow up when imitating a thing with generators on it

### LINKED ISSUE #273: Generator support
Up to this point, td.js has never had support for [ES generator functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*). This is a pretty large language feature, so I'm immediately intimidated of approaching it to naively. I'm not even sure yet how a generator user would _want_ it to work.

I've been reading a little bit from activity on our [sinon friends](https://github.com/sinonjs/sinon) repo, and found [this fork](https://github.com/ingameio/SinonES6.JS/commit/cf356ce29e8d36122e24cf1428cceefca8845184), which (I think) proposes simply wrapping a normal fake function in a generator, presumably so `next()` and `yield *` can be used, even if it just keeps returning whatever it's been stubbed to.

searls:
td.js does not yet support generator functions and we don't have a concept of faking / imitating them  since their stubbing needs are quite different from standard functions. However, as of the 3.2 release, anything users imitate that contain native generators will explode. This should fix that.

- [x] don't fake the generator with a test double to start with
- [ ] bail out when trying to change the constructor; either in chainPrototype (very deep) or in isPrimitiveLike (preventing the recurse step). 

The latter approach is cleaner and less likely to taint generators, but would prevent imitation of any user-defined properties on it. The former is more surgical in that it's possible we mess up something on the generators' prototype like next()

Fixes #273


searls:
My feeling is I should probably go with the isPrimitiveLike patch and rename it to something more intent-revealing like `isUnsupportedType()`

PR #282, opened 2017-08-18
Add previous/next links to documentation files

randycoulman:
This PR adds Previoius/Next links between the documentation files, making it easier to read the documentation end-to-end.

Along the way, I fixed a few inconsistencies and issues:
* The TOC in the README was not nesting properly.  It looks like GitHub requires the nested list number to be at the same indentation level as the text of the parent item, not just two spaces in.
* The TOC in the README was linking to the second-level Background header in `2-howto-purpose` instead of the top-level header.
* The TOC in the README was only linking to the `C-configuration` page, not the top-level header.
* `8-custom-matchers` was using level 3/4 headers instead of level 1/2 like all of the other documents.


searls:
Fantastic—thank you Randy!

PR #285, opened 2017-09-07
Upgrade lodash (and support async functions)

### LINKED ISSUE #284: Missing support for async functions? (Node 8)
td.replace and td.object do not seem to handle Node 8 async functions.

Example:

```
var fish = {
    eat: async function () { },
    swim: function () { },
    details: {
      age: 10,
      name: 'goldie'
    }
  };
  var fish = td.object(fish);
```

results of fish:
        { eat: {},
          swim: { [Function: testDouble] toString: [Function] },
          details: { age: 10, name: 'goldie' },
          toString: [Function] }

The async function eat is not a testDouble function - it gets replaced with a empty object.  I get the same results using td.replace().

rosston:
lodash@4.17.0 added support for async functions to `_.isFunction`, which allows testdouble.js to properly detect (mockable) functions on objects.

I believe this should fix #284.

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
   },
   "dependencies": {
     "es6-map": "^0.1.5",
-    "lodash": "^4.15.0",
+    "lodash": "^4.17.4",
     "quibble": "^0.5.1",
     "resolve": "^1.3.3",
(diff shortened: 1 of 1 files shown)


searls:
landed in 3.2.5

PR #299, opened 2017-11-11
WIP - remove dist from git

### LINKED ISSUE #290: Update build workflow to deprecate GitHub dist folder, instead publish through a CDN
testdouble.js is transpiled, minified and gzipped for delivery. However, then the release artifact is committed to the GitHub repo. Following the separation of concerns principle, the repo should only contain the source artifacts that are used to make up the final product. Therefore, I would like to propose deprecating the dist folder version of testdouble.js and instead publish library through a CDN.

Further:
- It is a best practice for a CI/build pipeline to manage release artifacts,
- It is annoying to create a commit that is not-related to a documentation or source artifact change,
- Without a CDN non-npm users loose human readable version information. The instructions asks users to download the latest directly from GitHub -- `$ curl https://raw.githubusercontent.com/testdouble/testdouble.js/master/dist/testdouble.js -o test/helpers/testdouble.js`. At download time, the user has no idea what version they're getting.

Suggested implementation: 
Unpkg seems to be able to easily host npm packages that implement a umd folder. See https://unpkg.com for further details.

Implementation would be fairly easy:
- [ ] Add the umd (or dist) directory to your .gitignore file
- [ ] Add the umd directory to your files array in package.json
- [ ] Use a build script to generate your UMD build in the umd directory when you publish
- [ ] Update documentation with the new CDN url
- [ ] Add deprecation notice to current dist file for a period of 3-6 months, followed by removal

jasonkarns:
closes #290 

npm runs `prepare` when a package is installed via git, so we just run compilation in the prepare script (which is incidentally the preferred script over prepublish)


searls:
Can you add some `- [ ]` to the comment to describe what's WIP about this? (other than the failing)

jasonkarns:
Eternally annoyed that browserify blows up when writing to a nonexistent directory instead of creating the necessary directories :/

Made `clean` script cross-platform (with mkdirp and rimraf devDeps). Compilation is safe after clean has been run.

jasonkarns:
Still testing yarn's compilation of git deps. Apparently it is quite broken:

https://github.com/yarnpkg/rfcs/pull/79
https://github.com/yarnpkg/yarn/pull/3553

jasonkarns:
Tests are failing because current "example" tests are broken (they're broken on master, too). Opening separate issue.

jasonkarns:
@searls I can get this cleaned up and ready to go if we're okay with one scenario not working:

yarn currently doesn't run the npm lifecycle scripts when installing deps from github. It _does_ when installing plain git deps, but yarn doesn't currently run the same install steps for github as they do for git.

https://github.com/yarnpkg/yarn/issues/5235
https://github.com/yarnpkg/rfcs/pull/79

So... if this PR were to be merged, the only people to be impacted would be yarn users installing from github. (yarn users installing from registry; and npm users installing from github would work as intended)

Are we okay with this? Personally, I am (unsurprisingly) since this is a yarn bug and I don't really feel compelled to have all the added annoyance of dealing with committing built assets to account for a bug.

searls:
Well, also users who download scripts from Github and then throw them into a vendor/js directory

jasonkarns:
We can just link to npmcdn for them.

searls:
Will merge if we can:

- [ ] rebase to latest master
- [ ] add links near the top of the README pointing to the npmcdm/latest
- [ ] see all builds pass

searls:
I swear to god I'm not trolling you but before we merge install directions also need to be updated. What would you have it say now, since most people are using npm/webpack for building browser apps whereas we were still living in bower/curl land when we started this thing?

https://github.com/testdouble/testdouble.js/blob/master/docs/1-installation.md#for-use-in-browsers

jasonkarns:
I can update those. I didn't even think to check the installation docs :(

Do we intend to drop bower support entirely or just undocument it? I can
look into what bower does with this setup.

PR #301, opened 2017-11-17
Symbols must be stringified explicitly

### LINKED ISSUE #124: td.callback, when given a td.object, raises a TypeError
This is on 1.6.0 of testdouble.js and 3.10.1 of lodash

Given this code:

```javascript
var td = require("testdouble");

var beingTested = {
  functionToTest: function(thing) {
    var self = this;
    thing.subscribe(
      function(response) {
        self.data = response.json().data;
      },
      function(error) {
      }
    );
  }
}

describe("", function() {
  it("", function() {
    var data = [
      {
        id: 1,
        name: "foo"
      },
      {
        id: 2,
        name: "bar"
      }
    ];
    var mockThing = td.object("subscribe");
    var mockResponse = td.object("json");
    /*
    var mockResponse = {
      json: td.function()
    }
    */
    td.when(mockThing.subscribe(td.callback(mockResponse),td.matchers.anything())).thenReturn();
    td.when(mockResponse.json()).thenReturn( { data: data });

    beingTested.functionToTest(mockThing);

    expect(beingTested.data).toBe(data);
  });
});
```

I get this output:

```
> node_modules/.bin/jasmine-node spec/javascripts/td.spec.js 
F

Failures:

  1)  
   Message:
     TypeError: Cannot convert a Symbol value to a string
   Stacktrace:
     TypeError: Cannot convert a Symbol value to a string
    at Object.Proxy.get (/Users/davec/Projects/rails-book/shine/node_modules/testdouble/lib/object.js:59:75)
    at Function.isString (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:8978:78)
    at module.exports (/Users/davec/Projects/rails-book/shine/node_modules/testdouble/lib/stringify/anything.js:10:11)
    at /Users/davec/Projects/rails-book/shine/node_modules/testdouble/lib/stringify/arguments.js:17:30
    at arrayMap (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:1406:25)
    at map (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:6710:14)
    at interceptor (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:12240:26)
    at thru (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:5927:26)
    at baseWrapperValue (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:2768:30)
    at LazyWrapper.lazyValue [as value] (/Users/davec/Projects/rails-book/shine/node_modules/lodash/index.js:1077:16)

Finished in 0.014 seconds
1 test, 1 assertion, 1 failure, 0 skipped


(node:8884) DeprecationWarning: util.print is deprecated. Use console.log instead.
```

If I replace

```javascript
 var mockResponse = td.object("json");
```

with
```javascript
    var mockResponse = {
      json: td.function()
    }
```

everything works.

I looked into the code, but can't understand what's going on enough to be more helpful :(

### LINKED ISSUE #300: Cannot convert a Symbol value to a string
When attempting to console log an object that has stubbed methods:

```
message: Cannot convert a Symbol value to a string
  stacktrace: TypeError: Cannot convert a Symbol value to a string
    at Object.get (/Users/jasonkarns/Projects/shutterstock/little-sister/node_modules/testdouble/lib/object.js:61:67)
    at formatValue (util.js:420:38)
    at formatProperty (util.js:828:11)
    at formatObject (util.js:644:17)
    at formatValue (util.js:606:18)
    at inspect (util.js:321:10)
    at format (util.js:188:12)
    at Console.log (console.js:127:21)
    at LittleSister.parseEventValue (/Users/jasonkarns/Projects/shutterstock/little-sister/src/index.js:84:11)
    at propsFromDataTrack (/Users/jasonkarns/Projects/shutterstock/little-sister/src/index.js:78:17)
```

I believe this is related to (or a dupe of #124)

## Investigation:

The error traces to: https://github.com/testdouble/testdouble.js/blob/v3.2.6/src/object.js#L38
which compiles as: https://github.com/testdouble/testdouble.js/blob/v3.2.6/dist/testdouble.js#L10675

According to https://github.com/nodejs/node/issues/927#issuecomment-75623216, implicit string conversion is not allowed on symbols.

jasonkarns:
Apparently (https://github.com/nodejs/node/issues/927#issuecomment-75623216),
Symbols cannot be implicitly coerced to string.

Template stringification leverages implicit coercion, so we must
stringify propKey before sending it to the template.

Options for explicitly stringifying:

- `String(propKey)`
- `String.prototype.toString(propKey)`

Differences:

```
> String(undefined)
'undefined'
> String.prototype.toString(undefined)
''
```

In this case, `propKey` represents the key for the stubbed function. If
the key is `undefined`, it would be implicitly coerced to `"undefined"`
as the property key. So `String()` is the explicit converstion we want.

fixes #300 and perhaps maybe fixes #124 ?

### DIFF
--- a/src/imitate/create-imitation.js
+++ b/src/imitate/create-imitation.js
@@ -11,7 +11,7 @@ export default (original, names) => {
       return original
     } else {
       // TODO: this will become src/function/create and include parent reference instead of name joining here
-      return tdFunction(names.join('') || '(anonymous function)')
+      return tdFunction(names.map(String).join('') || '(anonymous function)')
     }
   } else {
     return _.clone(original)
--- a/src/imitate/create-imitation.js
+++ b/src/imitate/create-imitation.js
@@ -11,7 +11,7 @@ export default (original, names) => {
       return original
     } else {
       // TODO: this will become src/function/create and include parent reference instead of name joining here
-      return tdFunction(names.map(String).join('') || '(anonymous function)')
+      return tdFunction(_.map(names, String).join('') || '(anonymous function)')
     }
   } else {
     return _.clone(original)
--- a/regression/src/constructor-test.coffee
+++ b/regression/src/constructor-test.coffee
@@ -72,6 +72,17 @@ describe 'td.constructor', ->
     And -> @fakeInstance.foo.toString() == '[test double for "#foo"]'
 
 
+  if (global.Symbol)
+    describe 'edge case: being given a Symbol as function name', ->
+      Given -> @symbolFoo = Symbol('foo')
+      Given -> @fakeConstructor = td.constructor([@symbolFoo])
+      Given -> @fakeInstance = new @fakeConstructor('biz')
+      Then -> @fakeConstructor.prototype[@symbolFoo] == @fakeInstance[@symbolFoo]
+      And -> td.verify(new @fakeConstructor('biz'))
+      And -> td.explain(@fakeInstance[@symbolFoo]).isTestDouble == true
+      And -> @fakeConstructor.toString() == '[test double for "(unnamed constructor)"]'
+      And -> @fakeInstance.toString() == '[test double instance of constructor]'
+      And -> @fakeInstance[@symbolFoo].toString() == '[test double for "#Symbol(foo)"]'
 
   describe 'edge case: being given a function without prototypal methods', ->
     Given -> @boringFunc = ->
--- a/regression/src/object-test.coffee
+++ b/regression/src/object-test.coffee
@@ -25,6 +25,15 @@ describe 'td.object', ->
     And -> @testDouble.toString() == '[test double object]'
     And -> @testDouble.bam.toString() == '[test double for ".bam"]'
 
+  if (global.Symbol)
+    context 'making a test double based on a Symbol', ->
+      Given -> @symbolFoo = Symbol('foo')
+      Given -> @testDouble = td.object([@symbolFoo])
+      When -> td.when(@testDouble[@symbolFoo]()).thenReturn('zing!')
+      Then -> @testDouble[@symbolFoo]() == 'zing!'
+      And -> @testDouble.toString() == '[test double object]'
+      And -> @testDouble[@symbolFoo].toString() == '[test double for ".Symbol(foo)"]'
+
   describe 'passing a function to td.object erroneously (1.x)', ->
     When -> try td.object(->) catch e then @result = e
     Then -> expect(@result.message).to.contain(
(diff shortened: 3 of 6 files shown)


searls:
Is there a test you could add for this?

jasonkarns:
tests in progress. Also found 3 other places where this bug can manifest:

- https://github.com/testdouble/testdouble.js/blob/fix-symbol-to-string/src/imitate/create-imitation.js#L14
- https://github.com/testdouble/testdouble.js/blob/fix-symbol-to-string/src/object.js#L29
- https://github.com/testdouble/testdouble.js/blob/fix-symbol-to-string/src/constructor.js#L16

So there are 4 total instances in all. Two of which use `.` as a dot-notation prefix for the method, and one uses `#` ala ruby's "method" prefix. Do we care to switch to `[Symbol(x)]` style function name, as that's the common idiom when using or describing Symbols-as-method-names? Without special handling, a symbol name could end up with a tdFunction name as:

`foo.Symbol(bar)`, or `#Symbol(bar)`, or just `.Symbol(bar)`

jasonkarns:
First commit is minimal change to fix the issue (just explicitly stringify the keynames)...

Second commit is an "improvement" to make the key names more idiomatic when stringified. (Symbol keys are wrapped in brackets). Not sure it's worth the extra code, though. Maybe something to revisit after the rewrite? I can nuke the second commit if we'd like to just do the minimal fix for now.

searls:
Is this done?

jasonkarns:
Yeah, it's ready if you're fine with the slight change to the stringified output of the function names (second commit)

searls:
Hey @jasonkarns, apologies that I just got around to reading this (mea culpa). My feeling is that since I'm still mid-rewrite (😭), I don't want to add any superfluous branching anywhere I can avoid it, so I'd really rather merge this fix in without any changes to the stringy output.

Could you update with changes that just make the fix without adding the `[…]` wrapper to symbol func names?

jasonkarns:
@searls done. there were already separate commits, so i just removed the latter one from the PR

src/imitate/create-imitation.js
```diff
@@ -11,7 +11,7 @@ export default (original, names) => {
       return original
     } else {
       // TODO: this will become src/function/create and include parent reference instead of name joining here
-      return tdFunction(names.join('') || '(anonymous function)')
+      return tdFunction(names.map(String).join('') || '(anonymous function)')
```
searls: Can you please use `_.map` for this? We're using lodash for all of our enumerable methods everywhere else, and there's a tacit (but not explicit) support for ancient browsers / IE etc that won't have `Array.prototype.map` defined.

jasonkarns:
@searls _.map 👍

searls:
Travis is having some downtime so I'm not going to watch this all day. How about you merge & release yourself, using this new guide I just wrote? https://github.com/testdouble/testdouble.js/blob/master/RELEASE.md

jasonkarns:
Ugh. `test:example` tests are failing, but they also fail at the merge-base with master :(

Probably gonna rebase to latest master, if that's green...

searls:
Definitely rebase to latest

PR #303, opened 2017-11-19
Fix example tests

### LINKED ISSUE #302: example tests don't work on latest master
Apparently, the node, webpack, and babel example directories have testdouble symlinked to the parent directory. Each of those test scripts run yarn-install and yarn-test within their respective example directories. Yarn install in each of those directories removes the symlink and installs testdouble from registry, thus defeating the point of the example tests.

jasonkarns:
fixes #302 

package.json for each of node, babel and webpack examples now uses relative path as version spec. yarn handles this by simply not installing testdouble into node_modules so it gets resolved by ancestor hierarchy. npm5 handles this by symlinking from node_modules to ancestor directory.

In both cases, testdouble dep is resolved as the live, in-repo module.


jasonkarns:
Looks like the lineman example is using a bundled build of testdouble? Unsure how that gets updated to point to latest builds...

searls:
IIRC, the lineman build depends on running browserify compilation and then symlinking to dist/testdouble.js

jasonkarns:
👍 didn't catch the symlink earlier. In that case, this PR should be good to go I think

searls:
Thanks Jason!

PR #310, opened 2017-12-10
fix: captors used in stubings should only capture if the whole stub spec is matched

### LINKED ISSUE #308: Disallow captors in `td.when(...)`
I would have expected the following test to fail:

```js
test('td', t => {
	const foo = td.function('foo');
	const captor = td.matchers.captor();

	td.when(foo(captor.capture(), 'baz')).thenReturn(undefined);

	foo('bar', 'bazzz'); // note the extra z's

	t.is(captor.value, 'bar');
});
```

However, it passes. Looking at the documentation, I quickly realized my mistake. I should have used `td.verify` after calling `foo`, instead of `t.when` before. In this contrived case, that actually ends up being less verbose since I avoid the `thenReturn` call, so it's definitely on me for not reading the docs carefully enough. 

That said, I think my usage is fairly intuitive, and seems like it should work. I got away with it for quite a few tests before realizing my mistake. Ideally, we should do something to help others avoid my mistake.

I propose one of two solutions:

 1. Make usage inside `when` captures an error, and inform the user: `"Captors may only be used in verify calls, not when calls"`.

2. Allow it, but the captor should only capture a value when the rest of the matchers match. In this case, the following should work:

    ```js
    foo('bar', 'bazzz');
    t.is(captor.values.length, 0); // not captured, because "bazzz" didn't match

    foo('bar', 'baz');
    t.is(captor.value, 'bar');
    ```

My preference would be to allow the second option, as it would allow you to cut down on verbosity when you actually do care about using `.thenReturn`. Also, if we choose the first option, you force users to combine `t.when`, and `t.verify` for the same interaction, [which you discourage](https://github.com/testdouble/testdouble.js/blob/master/docs/B-frequently-asked-questions.md#why-shouldnt-i-call-both-tdwhen-and-tdverify-for-a-single-interaction-with-a-test-double).

jamestalmage:
Fixes #308

### DIFF
--- a/regression/src/captor-test.coffee
+++ b/regression/src/captor-test.coffee
@@ -8,6 +8,17 @@ describe 'argument captors (a special sub-type of matchers)', ->
     Then -> @captor.value == "PANTS!"
     And -> @stubbing == 'foobaby'
 
+  describe 'when stubbing and other matchers do match', ->
+    Given -> td.when(@testDouble(@captor.capture(), 'PANTS!')).thenReturn('barbaby');
+    When -> @stubbing = @testDouble('SHIRTS!', 'PANTS!')
+    Then -> @captor.value == 'SHIRTS!'
+    And -> @stubbing == 'barbaby'
+
+  describe 'when stubbing and other matchers do not match', ->
+    Given -> td.when(@testDouble(@captor.capture(), 'PANTS!')).thenReturn('barbaby');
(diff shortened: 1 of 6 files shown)


searls:
Thanks James. Before merge I'm going to need to carefully re-read #308, the docs, this changeset, and think of counter-examples. It may be a little while. Really appreciate you offering this PR up.

searls:
Landed in 3.2.7

PR #311, opened 2017-12-20
Ensure captors only fire when invocation satisfies when/verify

### LINKED ISSUE #308: Disallow captors in `td.when(...)`

searls:
@jamestalmage made a good catch in #308 that captors could be used with `td.when` but it was undocumented and it would also capture values when the rest of the rehearsal did not match the actual args. This was effectively a requirements miss, but because our goal is to have `td.when` and `td.verify` to behave as symmetrically as possible, it's something we needed to fix.

James implemented a fix in #310, but I had three quibbles with it: (a) it broadened awareness of "capturing" to matchers themselves, even though captors are a one-off subset, (b) it expanded the private namespace claimed by matchers with `__capture`, and (c) it didn't verify that an argument was a matcher before invoking `__capture`.

- [x] Commit James' failing tests
- [x] implement an `afterSatisfaction` hook to the matcher creation, which only fires after all args match on an invocation or verification
- [x] document the new hook
- [x] make the new behavior work for the in-place rewrite that's still (slowly) underway; right now that means https://github.com/testdouble/testdouble.js/blob/master/src/when/index.js needs to be analyzed and updated with this responsibility

Fixes #310 
Fixes #308

### DIFF
--- a/docs/8-custom-matchers.md
+++ b/docs/8-custom-matchers.md
@@ -13,7 +13,7 @@ messages from `td.explain` calls and `td.verify` failures.
 
 (For the record, arguments without a `__matches` property will only match an
 actual invocation if it passes lodash's deep
-[_.isEqual](https://lodash.com/docs#isEqual) test.)
+[`_.isEqual`](https://lodash.com/docs#isEqual) test.)
 
 The examples in this document assume you've aliased `testdouble` to `td`.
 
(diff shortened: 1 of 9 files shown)


searls:
By the way, I've intentionally chosen not to document this `td.when` support for argument captors right now. If you're simply capturing a value (as opposed to a function), I think `argThat` is usually more succinct & local. If you're capturing a function, then `td.callback` is often more appropriate. I can't think of a place where I'd explicitly recommend it over other constructs provided by the lib

searls:
Also dependency CI appears to be broken. Disabling

searls:
Landed in 3.2.7

PR #317, opened 2017-12-25
allow using matchers inside contains

lgandecki:
This allows us to do things like:

```javascript
brew({nested: {nestedString: 'abc', irrelevant: true}, irrelevantHere: 'alsoTrue'})

td.verify(brew(td.matchers.contains({irrelevantHere: td.matchers.isA(String)})))
td.verify(brew(td.matchers.contains(
{nested: td.matchers.contains({nestedString: td.matchers.isA(String)})}
)));
```

Before, the tests I added were failing (basically, you weren't able to mix matcher.contains with other matchers (including another matcher.contains as showed above)

### DIFF
--- a/regression/src/matchers-test.coffee
+++ b/regression/src/matchers-test.coffee
@@ -94,6 +94,24 @@ describe '.matchers', ->
       Then -> @matches(td.matchers.contains(deep: {thing: 'stuff'}), {}) == false
       Then -> @matches(td.matchers.contains(deep: {thing: 'stuff'}), deep: {thing: 'stuff', shallow: 5}) == true
       Then -> @matches(td.matchers.contains({container: {size: 'S'}}), {ingredient: 'beans', container: { type: 'cup', size: 'S'}}) == true
+      Then -> @matches(td.matchers.contains({someString: td.matchers.isA(String)}), {someString: "beautifulString"}) == true
+      Then -> @matches(
+        td.matchers.contains({someString: td.matchers.isA(String)}),
+        {someString: "beautifulString", irrelevant: true}
+      ) == true
+      Then -> @matches(
+        td.matchers.contains({nested: {someString: td.matchers.isA(String)}, relevant: true}),
+        {nested: {someString: "beautifulString"}, relevant: true}
+      ) == true
+      Then -> @matches(
+        td.matchers.contains({someString: td.matchers.isA(String)}),
+        {someString: 4}
+      ) == false
+      Then -> @matches(
+        td.matchers.contains({nested: td.matchers.contains({nestedString: td.matchers.isA(String)})}),
+        {nested: {nestedString: "abc", irrelevant: true}, irrelevantHere: "alsoTrue"}
+      ) == true
+
 
     context 'regexp', ->
       Then -> @matches(td.matchers.contains(/abc/), 'abc') == true
--- a/src/matchers/builtin/contains.js
+++ b/src/matchers/builtin/contains.js
@@ -20,8 +20,15 @@ export default create({
   }
 })
 
-var containsAllSpecified = (containing, actual) =>
-  actual != null && _.every(containing, (val, key) =>
-    _.isObjectLike(val)
-      ? containsAllSpecified(val, actual[key])
-      : _.isEqual(val, actual[key]))
+var containsAllSpecified = (containing, actual) => {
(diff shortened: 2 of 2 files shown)


searls:
Very cool! I think @schoonology requested this a while back

searls:
I just added a couple additional checks to [this branch](https://github.com/testdouble/testdouble.js/tree/lgandecki-containsMatchers), and found one issue: that this does not apply to arrays as one would expect. Example test I added:

```js
this.matches(td.matchers.contains(td.matchers.isA(Number)), ['a', 3, 'b']) === true
```

https://github.com/testdouble/testdouble.js/blob/lgandecki-containsMatchers/regression/src/matchers-test.coffee#L91

I'll take a look to see if this is a trivial addition to the function you changed

searls:
I didn't have much luck making this behave consistently for arrays (the complexity of the private function is now at a point where it seems a single recursive function would be preferable to two half-functions given that `_.isEqual` checks will no longer cut it, but I didn't have time this morning to work out a good solution. I did refactor it slightly in 7e7b243e9866b895df16061a131ae25ac09c30a9 before quitting, though

lgandecki:
Thanks for the feedback! 
I don't really need the array option for the thing I'm working on currently, but I understand the need for consistency here. :)

I was able to make the tests pass with a bit extra complexity.. We could possibly split those in separate sub functions. I tried to break this but couldn't. My examples:

```javascript
const brewNew = td.function();
  brewNew(['sdf', {test: 'abc'}, 4, {testNew: 'sdf'}, {testAlso: {stringy: 'string', number: 4}}]);

  td.verify(
    brewNew(
      td.matchers.contains(
        {test: 'abc'},
        td.matchers.isA(String),
        td.matchers.isA(Number),
        {testNew: td.matchers.isA(String)},
        {
          testAlso: td.matchers.contains({
            stringy: 'string', number: td.matchers.isA(Number)
          }
          )
        }
      )
    )
  );
  const newer = td.function();

  newer({
    lol: {
      deep:
          [{ad: ['wer', {evenDeeper: [function () {}, 4]}, 3]}, 2]
    },
    other: 'stuff'
  });

  td.verify(newer(td.matchers.contains({
    lol: {
      deep: [{
        ad: ['wer', {evenDeeper: td.matchers.contains(td.matchers.isA(Number), 4)}]
      }]
    }
  })));
```

lgandecki:
I just realized that if I simplify the code more, to:

```
var containsAllSpecified = (containing, actual) => {
  if (_.isArray(actual)) {
    return _.some(actual, el => {
      if (isMatcher(containing)) {
        return containing.__matches(el)
      }
      return containsAllSpecified(containing, el)
    })
  }
  return actual != null && _.every(containing, (val, key) => {
    if (isMatcher(val)) {
      return val.__matches(actual[key])
    }
    if (_.isArray(containing)) {
      return containsAllSpecified(val, actual)
    }
    if (_.isObjectLike(val)) {
      return containsAllSpecified(val, actual[key])
    } else {
      return _.isEqual(val, actual[key])
    }
  })
}
```
(removing the  && !_.isArray(containing) check in the second line and wrapping the containing as an array)
I get the built-in tests to pass, but my safety net failed:


```
const newer = td.function();

  newer({
    lol: {
      deep:
          [{ad: ['wer', {evenDeeper: [function () {}, 4]}, 3]}, 2]
    },
    other: 'stuff'
  });

  td.verify(newer(td.matchers.contains({
    lol: {
      deep: [{
        ad: ['wer', {evenDeeper: td.matchers.contains(td.matchers.isA(Number), 4)}]
      }]
    }
  })));
```

So I've added a test to cover a more complex example like this one.  I switched from a function inside the array to an object - I couldn't figure out how to make something like that in coffeescript, and didn't particularly fancy learning that ;-)

lgandecki:
Hey, thanks for merging @searls ! The recent changes I talked about were here:
https://github.com/testdouble/testdouble.js/pull/318/files

Not sure if you decided it was making thet initially function too complex, or assumed my changes were here :-)

searls:
I iterated on this a bit more today and was able to simplify things quite a bit further. Just cut a release. Thanks for this contribution!!

Landed in 3.3.0.

lgandecki:
Right, I see now. Much cleaner - great work :-) Thanks for this fantastic library - it, and it's docs, changed the way I think about testing.

Issue #321, opened 2018-01-09
td.object(undefined) allowes to call any method


gotenxds:
Hi, I stumbled upon this bizarre behaviour,
Creating a test double with td.object with undefined as the object create a testdouble that will allow you to call anymethod.

```javascript
var td = require("testdouble")

x = td.object(undefined) // Has toString() but nothing else.

x.someMethod() // returns undefined.

x // Now has toString() and someMethod()
```

You  view this here 
https://runkit.com/gotenxds/5a5481a90fd28800158086f7

gurpreetatwal:
I believe this is expected behaviour

> If passed either a string name or no arguments at all, td.object will return an ES2015 Proxy object designed to forward any property access as if it were a test double function. By using Proxy, testdouble.js is able to intercept calls to properties that don't exist, immediately create a new test double function, and invoke that function for use in either stubbing or verifying behavior.


https://github.com/testdouble/testdouble.js/blob/master/docs/4-creating-test-doubles.md#objectobjectname

searls:
Indeed, this is intended behavior, but perhaps we should be counting our `arguments` and failing in this case, since there's a likely different in intent between `td.object()` and `td.object(undefined)`. If you're passing a reference to what you think is an object bag and it ends up being undefined, it's going to be pretty surprising when you're  handed an ES Proxy object.

I'd support a PR that blew up in this exact instance

gotenxds:
@searls it was pretty surprising 😂
I thought this might be intended but it was not expected and we discovered it by mistake as it does not fail any test.

searls:
Fix landed in 3.3.2. Thanks!

PR #322, opened 2018-01-10
Erroring on undefined for td.object

gotenxds:
As mentioned in #321


searls:
Landed in 3.3.2. Thanks!

PR #328, opened 2018-01-23
First draft of issue template

stevenjackson:
Just an idea/draft.

1.  Would an issue template make us more effective at handling issues?
1.  Is _this_ issue template a step in the right direction?


searls:
Will take a loom later. Thanks Steve, was going to ask someone to take a stab at this!

PR #342, opened 2018-02-21
Simplify build

searls:
Interested in experimenting with moving away from babel+browserify to typescript+rollup, but our build & devDeps are so absurdly complex that I threw in the towel last time. The goal of this PR is to significantly slim down our exposure to various devDeps and build steps by consolidating and eliminating stuff that the tdjs build is doing.

- [x] update deps (cherry-picked to master)
- [x] eliminate the browser-run of our unit tests, which means dropping testem, coffeeify, and headless chrome (may re-add later if it's easy under a new build system)
- [x] drop power-assert and its associated babel plugin (for now, may re-add later)
- [x] split our `test/unit` and `test/safe` tests, so that it's clear whether `td` means "known version of tdjs you can use to mock things" or "the current working copy being tested", respectively
- [x] migrate `regression/`'s test fixtures to `test/safe`
- [x] migrate `regression/typescript` to `test/typescript`
- [x] start migrating the actual coffeescript/mocha/GWT tests from `regression/src` to `test/safe`. This is a hugely manual pain in the ass, as one might expect.
- [x] add a teenytest plugin that lets us have a single combined unit test suite (simplyifying/fixing our coverage stats)


jasonkarns:
If you move off of browserify, you may also be able to eliminate the need to preemptively create the generated and dist directories. (Only browserify fails to create the necessary directory structure. babel, rollup and webpack all create any necessary directories themselves.) Assuming typescript is also that way, you can eliminate the postclean script, and thus also the mkdirp dep.

PR #343, opened 2018-02-25
Try out TypeScript

searls:
### Progress
- [x] install typescript
- [x] rip out babel
- [x] move our existing typings test in with the other tests, deleting its npm script & the generated directory
- [x] see `npm test` pass
- [x] verify the dist/testdouble.js looks okay
- [x] see the whole build pass
- [x] set up a second standard task for linting typescript
- [x] convert an actual JS file to typescript
- [x] measure build performance

### Blockers

- [x] Currently using `ts-node` for its require.extensions hook to run our tests, but it appears to blow up on Node v4 which is failing the build. (Fix: the explosion was caused by me targeting es2015 instead of es5)


searls:
I'm deciding to merge for two reasons:

1. It seems to mostly work
2. It's much faster than under babel. 

In particular, `$ time npm t` takes this long under Node v8.90 on master:

```
real	0m9.226s
user	0m7.091s
sys	0m0.998s
```

And here's the same command compiled under typescript:

```
real	0m1.945s
user 0m1.566s
sys	0m0.189s
```

For a better than 4x speedup in walltime for my test feedback loop, I'd gladly switch compilers even if I never write a single '.ts' file.

searls:
Okay, and I figured out why `standard` had been taking 30s. Turns out its `deglob` dependency redundantly ends up traversing ignored directories unless they're explicitly ignored like "examples/**", which cut our lint time from 30s to 1s.

The time for `npm run test:ci` is now down to:

```
real	0m35.419s
user	1m12.684s
sys	0m11.266s
```

PR #344, opened 2018-02-25
Speed up the build

searls:
Try to make `npm run test:ci` faster by sprinkling in some parallel tasks, particularly (a) the styling + the tests and (b) each of the example projects. All results local.

master:
```
real	1m21.797s
user	1m18.127s
sys	0m15.781s
```

with npm-run-all's `run-p`:
```
real	0m53.442s
user	1m42.984s
sys	0m18.903s
```


searls:
The differences on travis were pretty dramatic despite small sample size.

This branch:
screen shot 2018-02-25 at 16 32 57

Master:
screen shot 2018-02-25 at 16 32 32

Master is an outlier, but this patch would probably reduce variability by running things parallel rather than serially

searls:
Okay, and I figured out why `standard` had been taking 30s. Turns out its `deglob` dependency redundantly ends up traversing ignored directories unless they're explicitly ignored like "examples/**", which cut our lint time from 30s to 1s.

The time for `npm run test:ci` is now down to:

```
real	0m35.419s
user	1m12.684s
sys	0m11.266s
```

searls:
Each travis build was also a good bit faster:
screen shot 2018-02-25 at 17 11 02

searls:
Going to merge now. Cache slowed build down but we should watch it over 2 or 3 builds on master to see if that was a one time cost

searls:
To close the loop, after warming the cache on master travis is indeed faster with the various node_modules directories cached
screen shot 2018-02-25 at 19 48 25

PR #349, opened 2018-02-27
Merge Circle CI config to master

searls:
- [x] rebase if needed
- [x] see travis still passing
- [x] see circle passing

PR #350, opened 2018-02-27
Re-enable caching for npm

zzak:
cc #349

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -4,7 +4,7 @@
 
 export type DoubledObject < T > = T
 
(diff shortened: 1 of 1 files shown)


zzak:
@searls hmm looks like it failed on an older node, i wonder if it's because it shared the cache for the same versions of node/npm would be bad?

I'll have to take a look at this later if you don't figure it out.

Another thing I'd like to do is get [test reporting](https://circleci.com/docs/2.0/collect-test-data/) setup, we can show test failures and some graphs but requires outputting results to junit and storing the xml some place.

Issue #168, opened 2016-12-17
testdouble causes bluebird to emit Unhandled rejection


queston02:
When we use _td.matchers.anything()_ with _thenReject_ bluebird starts emiting "Unhandled rejection" message in console. Below is an example ...

td.when(someService.doSomething(td.matchers.anything())).thenReject('Some Error');

If in the above code i replace _td.matchers.anything()_ with exact parameter then the messages stops. I think testdouble is not able to notify bluebird that promise rejection is handled when we are using _td.matchers.anything()_  

Let me know if you need full code from me to reproduce this.

searls:
A couple thoughts here:

1. I think going forward for edge case questions like this, adding a failing example to the `example/` directory (to an existing project if possible, new project if necessary), will really help us a lot to save on up-front communication about what the problem is. In this case, you should be able to add a failing example to the `example/node` project by installing bluebird there.

2. Yeah, what you're describing sounds broken, but I don't use bluebird a whole lot.

3. @schoonology, when you talked about a default issue template, I'd 100% support that if it also asked for an example test in the case of bugs/unexpected behavior.

terinjokes:
I'm using the native Promises in recent versions of Node.js and get a similar warning.

```
(node:3358) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): undefined
(node:3358) DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

I'm not sure if this is a bad pattern, but I've recently been wiring things up with code similar to the following.

```javascript
td.when(getter.got(), {ignoreExtraArgs: true}).thenReject();
td.when(getter.got('path/to/something').thenResolve(body);
```

Otherwise, the test failures were pointing me in the wrong direction (complaining about the subject calling `.then` on undefined) when not implemented correctly.

cjthompson:
I'm having an issue with `thenReject` also causing an unhandled rejection. 

The code is pretty straight forward. 

```
class Cache {
  function check() {
    return status.getDeployDate()
      .then(...)
      .catch((e) => this.emit('error'));
  }
}
```
and the test
```
  const status = td.replace('../../../src/lib/status');
  td.when(status.getDeployDate()).thenReject(new Error());
  const cache  = new Cache();
  cache.on('error', (e) => assert(e.message, 'error'));
  cache().then(() => done());
```

[...]

PR #385, opened 2018-10-08
feat: Add ava tests examples

pke:
Adds ava nodejs tests based on the node tests.

It should showcase how to write tests for ava with td.js (see #168)


pke:
@searls seems like an outdated jsdom dependency makes the tests fail?

examples/node-ava/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "testdouble-nodejs-example",
```
searls: let's name this `testdouble-ava-example`

examples/node-ava/test/lib/instance-of-test.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+class EightiesGuy extends Person {
+  hairspray () {
+    return true
+  }
+}
+
+test.beforeEach(function (t) {
+  t.context.td = require('../../../..')
```
searls: Is this relative path necessary? The `node/` example in the same repo is able to just do `global.td = require('testdouble')` by naming `"testdouble": "file:../.."` as a devDep in its package.json

Also, I don't think we should assign `td` in a beforeEach. It can happen once and should be safe at the top of a shared test helper and assigned globally.

searls:
Try rebasing master now that #386 has landed

examples/node-ava/test/lib/instance-of-test.js
pke: On my other machine I was not able to run `npm install` with the file dependency. It always errored with some error in the `.cache` folder. However on another machine it runs fine (maybe --no-package-lock) made the difference or a differen node/npm version. I'll check later.

searls: there are lots of issues pertaining to `file:` deps and package-lock. since this is just an example project, nuke the package lock if you need to and get it committed in a working state

pke:
There @searls all green now :)

searls:
the more the merrier!

Issue #170, opened 2016-12-18
td.verify and td.when redundency warning


LandonSchropp:
I have a test suite that periodically invokes a stubbed function on an object. and consumes the result. Since that function gets reused repeatedly in my test suite, I've defined it in a before each block for all of the specs.

However, in one particular spec I want to verify the test double *wasn't* called. Unfortunately, I get a warning that both stubbing and verifying the call with the same arguments is redundant. This warning makes perfect sense, but I don't think it's applicable to situations where `times` is anything other than `1`.

Here's a simplified example:

```
describe("example", () => {
  let func;

  beforeEach(() => {
    func = td.function();
    td.when(func()).thenReturn("Nope");
  });

  it('is never called', () => {
    td.verify(func(), { times: 0 });
  });
});
```

And this is the warning:

```
Warning: testdouble.js - td.verify - test double was both stubbed and verified with arguments (), which is redundant and probably unnecessary. (see: https://github.com/testdouble/testdouble.js/blob/master/docs/B-frequently-asked-questions.md#why-shouldnt-i-call-both-tdwhen-and-tdverify-for-a-single-interaction-with-a-test-double )
```

searls:
There's two issues at play here:

* Configuration in cascading test-styles like the nested example groups in Jasmine or Mocha mean that a generic test double configuration should be able to be made more and more specific as you get closer to a specific example. This works pretty well with stubbings, because last-in-satisfied-stubbing-wins.
* When you have a stubbing that's called under conditions `A`, `B`, and `C`, but the test author knows it won't need to be called at `D`, the question arises "what should the test do here"?

I would argue that verifying a stubbing wasn't called is _probably_ still not necessary in this case. To recap: `td.verify` should only be used if there's no way to verify the subject did something, which is typically when the test double function has a side effect and doesn't return a value. 

I'll expand your example below:

```js
describe("example", () => {
  let func;

  beforeEach(() => {
    func = td.function();
    td.when(func()).thenReturn("Nope");
  });

  it('is called when A', () => {
    var result = subject('A', func)

    expect(result).to.equal('pants')
  });

  it('is never called when B', () => {
    var result subject('B', func)

    expect(result).to.equal('cookie')
    // td.verify(func(), { times: 0 }); // <- this would be redundant and over-prescriptive.
  });
});
```

[...]

PR #391, opened 2018-11-16
Add explanation for verifying stub was not called

brandonduff:
I often find myself referring to https://github.com/testdouble/testdouble.js/issues/170#issuecomment-268571603 in discussions about this topic with colleagues. I figured it may be a common enough question that it would be useful to have it listed in the docs. I basically just copied the comment over.


searls:
Thank you for submitting this Brandon. I pulled the code down and spent nearly an hour trying to rework it, but all I managed to do was unconvince myself that this is good across-the-board advice. I think the question above it (about stubbing & verifying the same interaction) is more salient. I wasn't able to find a way to word this that I agree with in 90% of cases (realizing that yes, this means I'm disagreeing with my past self somewhat). Honestly it has me thinking this warning might be overly prescriptive.

Going to close this for now and chew on it further.

brandonduff:
That makes sense to me. This does seem like a case with a lot of exceptions. The most general form of this advice I try to give is something like "focus on specifying how your subject behaves; not how it doesn't behave." But you're right, this is not as widely-applicable as the above question.

Thanks for thinking about this! I'd be curious to hear what you come up with on this subject.

Issue #377, opened 2018-06-15
Add sticky `td.reset.onReset` hook


searls:
[testdouble-nock](https://github.com/testdouble/testdouble-nock) would be a lot nicer if we had a sticky `onReset` hook, not just a single use `onNextReset`, so that the library can configure a reset on the end-user of a plugin's behalf

searls:
Fixed by #393

PR #393, opened 2018-11-19
Add onReset hook.

primarilysnark:
Pursuant to #377, this PR adds support for an `onReset` hook. It also adds a test spec for reset functionality as I noticed one was missing.


searls:
This passes and looks fine so I'm going to merge and cut a release before I head out on vacation, but we should really probably document both of these in docs/…

searls:
landed in testdouble@3.9.0 -- thanks Josh!

Issue #395, opened 2018-11-28
TypeScript: Update Definition


lisamai:
### Description
Hi, I was wondering if there was any plan to update the TypeScript definitions for the project? Since the last update I think quibble and imitate have been updated to the API. Happy to help out and submit a PR for this as well.

Wasn't sure whether to open a new issue since #238 is still open but that was last year and the associated PR was merged.

Thanks!

PR #396, opened 2018-11-29
Add type information for imitation to the exported typings file

lisamai:
Added imitation function to the exported types file for #395. After looking through, I think it's just imitate that was added in the public API doc. I'm not too familiar with the underlying code here, so feel free to nitpick away.

I also updated the minimum version for npm-run-all to 4.1.5 due to mystictea/npm-run-all#149 / dominictarr/event-stream#116 I can break out this section if desired, but included it as I needed to do this for `npm install` to succeed.


searls:
Thanks! Landed in 3.9.1

PR #398, opened 2018-12-21
Explain object

robwold:
Well, you said you wanted a dirty proof of concept ...

`td.explain(myObject)` produces something with basic `description` and `isTestDouble` fields, together with a field for each method of `myObject` that's the result of passing that method to the old `td.explain` (now a private method called `explainFunction`). Nested objects are handled recursively, and this structure is reflected in the output of `td.explain`.

Some tests of the new behaviour have been added to `test/safe/explain.test.js`. (I'm not entirely sure of the intention of a 'safe' test directory tbh, but this seemed like the place to put them?)

Note also that lots of versions got bumped when I ran `npm install`, and `npm audit` still isn't entirely happy.

Happy to keep plugging at this if this is needs to be either extended or re-worked totally.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,6 @@ npm-debug.log*
 
 .node-version
 .vscode
+
+# For webstorm IDE
+./idea/*
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,3 @@ npm-debug.log*
 
 .node-version
 .vscode
-
-# For webstorm IDE
-./idea/*
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,10 +20,10 @@
       "dev": true,
       "requires": {
         "@babel/types": "7.0.0-beta.51",
-        "jsesc": "2.5.1",
-        "lodash": "4.17.11",
-        "source-map": "0.5.7",
-        "trim-right": "1.0.1"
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -14,3 +14,6 @@ npm-debug.log*
 
 .node-version
 .vscode
+
+# For webstorm IDE
+./idea/*
```
searls: Please use your own [global excludes](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) instead of adding them to the project

searls:
Thanks for this! I'm sorry I won't have time to review this over the holidays but I hope to get to it in the next week or two. Feel free to remind me if I don't!

.gitignore
robwold: Done. This felt kinda crummy, thanks for the pointer.

robwold:
> Thanks for this! I'm sorry I won't have time to review this over the holidays but I hope to get to it in the next week or two. Feel free to remind me if I don't!

@searls Just thought I'd give this a bump.

searls:
I have released this feature in 3.10.0

See changes in for an illustration on where I took this: https://github.com/testdouble/testdouble.js/blob/master/test/safe/explain.test.js

Issue #356, opened 2018-03-22
How to mock typescript abstract class?


stepan-prokipchyn:
### Description
I'm trying to mock abstract class without implementation:
```js
it("should call dismiss when close is clicked", () => 
    {
        var notificationService = td.object(INotificationService.prototype);

        var component = new NotificationComponent(notificationService);
        component.reportCloseClicked();

        td.verify(notificationService.dismissRequested());
    });
});
```
### Issue

'dismissRequested' is not found in 'notificationService'

stepan-prokipchyn:
```js
import { Observable } from 'rxjs/Observable';

/**
 * @interface INotificationService
 * 
 * @brief Interface for classes which handles notification behaviour.
 * 
 * This interface hides specific notification behaviour behind standard actions.
 */
export abstract class INotificationService {

    /**
     * @brief Called when user presses dismiss button.
     * 
     * Notification component might be hidden by implementation.
     */
    abstract dismissRequested();

    /**
     * @brief Called when user presses "More information" button.
     * 
     * Usually new page or popup will be shown.
     */
    abstract detailsRequested();
};
```

searls:
This is just a guess, but you probably want your test to use `td.constructor`, not `td.object`.

```
var NotificationService = td.constructor(INotificationService);
```

The fake `NotificationService` that results can then be instantiated by your test or your subject

stepan-prokipchyn:
@searls thank you for suggestion. However I'm getting the following error in your code:
`Argument of type 'typeof INotificationService' is not assignable to parameter of type 'Constructor<INotificationService>'.`
Do you think a test project which reproduces the issue might help?

searls:
Sure, but I have to admit I'm hardly a typescript expert.

stepan-prokipchyn:
@searls , I ended up with the following approach that works for me:
`var notificationService = <INotificationService>td.object(["dismissRequested"]);`
So, I have to mock some anonymous object with all methods I need and then cast it to my base class. From one side this is more verbose, but from other side - it's more clear which methods are used.

What do you think about that?

I've also created a question about how to get a list of abstract methods to mock them automatically:
https://stackoverflow.com/questions/49487176/is-it-possible-to-iterate-typescript-class-abstract-methods-in-run-time

searls:
Sorry, I'm not familiar enough with TypeScript to comment. Paging @duluca in case he has time

duluca:
[...]

PR #399, opened 2018-12-25
typeof function for interfaces

lgandecki:
As per discussion in #356 

This is the initial work for getting testdouble work with interfaces.
Missing:
[] Working tests
[] Documentation

I'm not sure how to write a nice test for it (we are getting so meta..) - I hope that for someone that knows the codebase better (and what is going on behind the tinyteest setup for tests) it will be obvious what I'm doing wrong. The same code works just fine if I import the linked testdouble and try it in actual tests. But, I'm not even sure if that's the best way to test this - so I didn't want to spend too much time digging around.

To show you how beautiful workflow this allows:

screen shot 2018-12-25 at 9 32 11 am

with all the typescript goodies:

screen shot 2018-12-25 at 9 32 28 am
screen shot 2018-12-25 at 9 32 16 am

If this looks good and we can figure out what to do with tests, I'm happy to write some docs (unless Justin wants to do so - the docs here are very strongly opinionated after all, and go much beyond showing the API) 

I would also like to take the next step here and talk about typing the .thenReturn and thenResolve like here:

screen shot 2018-12-25 at 10 10 45 am

In real life situation I made this kind of a helper for myself and then I get this help:

screen shot 2018-12-25 at 10 11 50 am

Not only the IDE can catch typos (which could possibly lead to tests passing but app not working in production), but it can also jump to the property definition (the blue dot with red arrow next to the line number).

If this is something that sounds reasonable, I'm happy to create a PR as well.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -254,3 +254,5 @@ export function when(...args: any[]): Stubber;
  * @returns {Stubber}
  */
 export function verify(a: any, check?: VerificationConfig): void;
+
+export function typeOf<T>(): T;
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,7 @@ import config from './config'
 import callback from './callback'
 import version from './version'
 import * as quibble from 'quibble'
+import typeOf from './typeOf'
 
 module.exports = {
   function: tdFunction,
@@ -19,6 +20,7 @@ module.exports = {
   object,
   constructor,
   imitate,
+  typeOf,
   when,
   verify,
   matchers,
--- a/src/typeOf/index.js
+++ b/src/typeOf/index.js
@@ -1,0 +1,16 @@
+import func from '../function'
+
+export default function typeOf () {
+  const mocks = {}
+
+  const get = property => {
+    if (!mocks[property]) {
+      mocks[property] = func()
+    }
+    return mocks[property]
+  }
+
+  return new Proxy(() => {}, {
+    get: (_, property) => get(property)
+  })
+}
--- a/test/unit/index.test.js
+++ b/test/unit/index.test.js
@@ -7,6 +7,7 @@ module.exports = () => {
   const constructor = td.replace('../../src/constructor').default
   const replace = td.replace('../../src/replace').default
   const imitate = td.replace('../../src/imitate').default
+  const typeOf = td.replace('../../src/typeOf').default
   // Stubbing & Verifying
   const when = td.replace('../../src/when').default
   const verify = td.replace('../../src/verify').default
@@ -28,6 +29,7 @@ module.exports = () => {
     constructor: constructor,
     replace: replace,
     imitate: imitate,
+    typeOf: typeOf,
     when: when,
     verify: verify,
     matchers: matchers,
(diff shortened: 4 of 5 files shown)


wpannell:
@searls this is interesting.

@lgandecki I would put your `typeOf` feature to good use.  It seems you are close except for the 2 failing tests.

I'm not sure what you are trying to demonstrate with those tests.  There does not seem to be a counterpart elsewhere in the codebase.

Perhaps the test for [`td.constructor`](https://github.com/testdouble/testdouble.js/blob/master/test/safe/constructor.test.js) is closer to what you need to prove (although, I'm no expert)?

I would recommend you submit this as a pull request and let the contributors help you prove your feature is sound.

searls:
Forgive me that I still haven't had time to review this patch, but a couple things stood out to me about the implementation.

1. What differs between this and the `td.object(['string name'])` API ([doc](https://github.com/testdouble/testdouble.js/blob/master/docs/4-creating-test-doubles.md#objectobjectname)), which also returns a `Proxy` instance? 

2. Along similar lines, because this doesn't call thru to `td.imitate()` (which our other fake-creators do), it means things like static properties won't get picked up and will be treated like functions, right?

searls:
@lgandecki to be clear, `td.object([string name])` does not look at a real type, it creates a proxy like this does. I just wanted to know what the difference was between this feature and that one

lgandecki:
@searls not sure what you are saying. The difference is in the result, this typeOf gives you a typed testdouble based on an interface.. you can use the.object but you won't get any typesafety so you could mock things that are not correct, get your tests to pass even though it won't work in production. 
Are you saying that it would be better to use the td.object internally? Might be, I didn't look at the internals of that to be honest. I was looking for a way to mock based on an interface and saw the ongoing discussion #356 .

searls:
Yes, I am suggesting that it makes more sense to me that this feature be added to our existing ways to create fake things, and because `td.object` already returns a `Proxy`, whatever is needed to add type information to this API should be applied to `td.object`'s Proxy API as a starting point (and then hopefully spread to other methods).

Where I'm coming from is that we already have way too many methods that create test doubles, which is both confusing to users and hard to maintain. On the maintenance, my main goal is to reduce the number of code paths each of them (i.e. `td.func`, `td.object`, `td.constructor`) call through internally

lgandecki:
Got ya! Makes sense! I definitely agree that the fewer functions td have the better - I was going off the idea from the mentioned issue where the suggestion was to create a new API.  I will take a look at this tomorrow, but it should be rather easy.

Thanks for the feedback!

wpannell:
I'm not sure of the utility of statics for getters/setters for test driving protocols between collaborators.

Perhaps it's possible to simplify and deliver a caveat experiment with the feature described above?

wpannell:
I'm following this thread with great interest.  I've got a group of newbies training right now, test-driving angular 7, and getting a lot of mileage out of `testdouble.js`.  This would be a great tool in their toolbox and you would have a whole group of us as early adopters for feedback.

lgandecki:
continued in #400

PR #400, opened 2018-12-29
allow for creating testdoubles using interface only

lgandecki:
continuing from #399 
@searls was right, the object function has all the necessary functionality built in. The only thing I had to tweak was this new type that allows for:
```javascript
td.object<IMyInterface>(); 
```
but even with the current code testdouble actually allowed for mocking based on an interface, but with a bit of redundancy:
```javascript
td.object<IMyInterface>("");
````

@wpannell you can use the second one already (with an argument of a string) for your training, with this merged in you will be able to remove the argument for a bit cleaner syntax.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -167,6 +167,7 @@ export function object<T extends string>(props: T[]): DoubledObjectWithKey<T>;
  * @returns {DoubledObject<T>}
  */
 export function object<T>(object: string): DoubledObject<T>;
+export function object<T>(): DoubledObject<T>;
 
 /**
  * Create a fake object that is deep copy of the given object.


wpannell:
@lgandecki awesome

searls:
Yep, seems legit. @lgandecki since I'm not really a TS user, would you mind opening a follow-on PR that does a couple things?

1. Documents this. I imagine a brief description in README and a `docs/10-using-with-typescript.md` file with examples or screenshots would really be great

2. Add typings to other top-level `td` APIs that might be better typified. Seems odd only `td.object` would do it.

lgandecki:
Sure!
I'm off this week starting in a few hours, and I promised myself not to do anything programming-related... I will jump on it right away when I'm back. 

Happy new year!

searls:
Landed in 3.9.2

lgandecki:
I've attempted something for 1. in #401 which to some degree also covers 2.
We could still make it so td.imitate<> works with its magic power, but it would require a little bit of hacking in the code itself, besides just changing the typings. I might give it a try at some point.

PR #401, opened 2019-01-11
Typescript usage docs

lgandecki:
Hey @searls , please take a look :)
Writing documentation to someone elses project always makes me a bit uneasy.. :) The PR is just a suggestion, feel free to rip it apart and tell me how differently would you like it, or if you prefer - go ahead and change whatever you want.
I won't mind, worst case scenario I will reuse parts of what I've done (especially around the example) to create a blog post.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -65,7 +65,6 @@ section:
     1. [Interfaces](10-using-with-typescript.md#interfaces)
     2. [Function Types](10-using-with-typescript.md#function-types)
     3. [Abstract and concrete classes](10-using-with-typescript.md#abstract-and-concrete-classes)
-    4. [On reasoning behind using abstractions](10-using-with-typescript.md#on-reasoning-behind-using-abstractions)
 11. [Plugins](A-plugins.md#plugins)
     1. [testdouble-chai](https://github.com/basecase/testdouble-chai)
     2. [testdouble-jasmine](https://github.com/BrianGenisio/testdouble-jasmine)


searls:
Hey @lgandecki thanks a bunch for getting this rolling. Looks good.

Only critical feedback I have is that I'd rather avoid the discussion in section "On reasoning behind using abstractions" in this lib's docs, because I disagree with the position. Would you mind replacing that section with a link (to a post of your own, to another post on the topic) in the section on concrete classes that just calls out "some TDD practitioners advocate mocking abstract interfaces instead of class implementations"?

If you're interested in my reasoning, I started hearing this advice when using Java Spring in 2005 (when the library literally wouldn't let you inject a non-interface), and again when various Java mocking libraries rose in prominence (and wouldn't let you mock classes), but in practice the advice was always counterproductive. Grepping for files in a project, you'd basically always run into exactly one implementation for every interface, and developers would very quickly fall into the habit of mindlessly copy-pasting what they need from an interface into an implementation. It was like littering a project with C header files but without the technical benefit.

I'm not saying the practice can't be valuable to teams who really internalize and embrace the spirit behind it, but I believe it's counter-productive as general advice.

lgandecki:
No problem! I can just remove that part altogether. It was a long shot  to put it there - the docs are highly opinionated, following YOUR ideas, and I (and many people I shared them with) definitely love them for that. :) 

What do you feel about that gif there? I think it gives a good feel about the possible workflow and speaks "a thousand words" (or a million, since it's a video instead of a picture).
I could just leave a section that says something more along the line of "Example of an interface-based workflow" but skipping all the discussion.

On the other hand - I would definitely like to learn more here. Are you saying you don't agree with the Dependency Inversion principle in general? Because of the increased ratio of noise to signal? 
I'm curious what you mean by the "mindless copy-pasting" from interface into an implementation - now IDEs do that for us, but how else would you do it, in a productive way (embracing the spirit behind it, as you said). 
I'm coming from more of a Ruby/Python/JS background. Only in the last few years doing more with languages like C#, Java, Kotlin, so I'm learning, going through the "classics".. 

Anything you would recommend to read or watch in this topic?

Thanks!

marcpeabody:
Thank you so much for being a "I would definitely like to learn more here" developer @lgandecki and for making these contributions. I can tell from this PR alone that you are the kind of of person I'd like to work with.

Maybe I can jump in on the DI question here. I can remember having discussions with Searls about this ages ago. It is something I ranted about a lot back then and was actually a huge influence on my decision to leave my Java career for Ruby.

Here's my grumpy old man history to give some context about how some of us feel badly about interfaces-with-one-implementation. In no way am I claiming to be correct or even credible. This is just my salty, grumpy take on it. I think Searls' position is pretty similar.

In our past we heard a programming principle "program to the interface" which, in Java, became very confusing to us all because we had a keyword literally called "interface". The principle was really about having an API mindset while designing every unit and exposing only the necessary parts. You can totally accomplish this without ever using a literal Java "interface" definition. But we started having projects where people would make an interface file for every class for no good reason other than they thought it was good practice and it got way worse when Spring came around. I'll come back to that.

Before we had Spring there was a lot of room for structuring Java projects however we wanted. As the needs of "Enterprise" applications grew, it left us with some project structure anarchy. The flow of method calls from class to class can make a confusing [digraph](https://en.wikipedia.org/wiki/Directed_graph) and we really needed more structure.

Our savior at the time was [J2EE Patterns](http://www.corej2eepatterns.com/). Virtually every Java project was using many of these patterns. It was even quite common to put the pattern name in the class name. Usually each group of patterns were in their own respective package directories. It was easy, at a glance, to know where to find virtually anything and the responsibility of each class. You could leave one Java project and go to another and there would be hardly any ramp-up time to learn the codebase.

The unfortunate side effect of these patterns was that smaller Java projects (ie simple CRUD) didn't need very many of the patterns but the Java Architects said we needed all of them all of the time. So nearly every Java project was using SessionFacades and BusinessDelegates and the like but our implementations were often just pass-through classes like an over-engineered bucket brigade handing our data (TransferObjects) from the views to the DB and back again. We'd also make every project a "three tiered" solution (note the three colors of the patterns diagram) just in case we'd ever want to literally move the backend code to a separate server. Just in case.

IIRC, this is around the time Rails came out. The idea of just models, views, and controllers to build a simple app was a very refreshing idea. Most of the Java apps being built were glorified CRUD apps anyway. We even had a pretty decent framework called Struts that popularized the MVC concept for the web. But we still had to use all the patterns too.

[Take a quick break and watch this metaphor for how Java Enterprise grew.](https://www.youtube.com/watch?v=sh-gB07JftY)

Then Hibernate (and other ORMs) started taking over in Java. It was a huge win over the older version of EJBs for DB strategy. It felt more natural to developers than the old EJB and was usually way faster to develop. Somewhere in that time period of people changing their DB strategies from EJB to Hibernate, Spring slipped in.

Java recruiters ceased looking for resumes with Java exam certifications and shifted their searches to "Hibernate" and "Spring".

Spring (the over-arching DI framework part of it at least) became the crutch for everyone wanting to switch to Hibernate but wary that they'd want to switch the "third tier" of their application to yet another technology again someday.

With Spring, you also never have to say "new SuperCoolSessionFacade()" ever again. It didn't really save any lines of code, but we were starting to hear that "new" was bad and how great it would be to swap in an UltraSuperCoolSessionFacade instance without having to rebuild the entire app. The swap-out also sounded cool for your test environments to mock or stub out part of your app - like your service calls.

FTR, I don't recall Spring requiring the use of literal "interface" types unless you actually had more than one implementation. I thought you could just inject the one class you have into another. Perhaps in the earliest versions (Searls mentioned 2005) you had to. I don't know.

Anyway, for some reason nearly every Java project started using Spring and along with it a separate "interface" file was getting defined for every class instance for pretty much no reason. You would hear Java developers talk strategy over lunch at tech conferences for how to do this consistently. You might call it SuperCoolSessionFacadeInterface so it showed up in your file list next to its one and only implementation. Some people made separate package directories for the interfaces so they didn't get in the way.

I would argue with my team that our IDE has a neato new tool to extract an "interface" if the need arises for an alternate implementation. We don't need an "interface" for every class and this isn't what "program to the interface" means. I never won that argument. Ever.

If there are two possible solutions, I want the one that is simplest. Usually "simplest" is also less code but not always. Usually it's the bare minimum of what is needed today but sometimes it means adding something so maintenance is easier for the next person. It's a balance. Do the thing that will make the world happier.

DI isn't definitively good or evil, but there were things that its introduction brought to some of our careers that were frustrating. Finding our projects flooded with pointless singly-implemented interfaces was one of them for me. It may sound trivial but at the time it was a tiny straw that broke the camel's back for Java Enterprise becoming overly complicated and I left to become a Rails contractor.

If you ever feel the need to create something like a literal "interface" for a single implementation, please just ask yourself why. In what future scenario are you or another developer going to look back and say "Gosh, I'm so glad this was there because my life is so much easier now!"?

Ultimately, choosing one way or the other really isn't the end of the world. And you'll find friends on both sides of the issue. Personally, I feel happier on projects without unnecessary files or definitions to traverse and I feel like interfaces with only one implementation usually add more clutter than clarity. I think Searls feels the same way and he wants to encourage other developers to follow the practices that make him happy because he hopes it will also make them happy.

samhatoum:
@marcpeabody thank you for replaying my former Java life and why I'm so glad to no longer be there anymore.

@lgandecki 
> Are you saying you don't agree with the Dependency Inversion principle in general?

I don't think DI is the issue here, it's more about disagreeing with interfaces that end up with just one implementation, and that a test-double library should be agnostic to the style that developers choose to do DI, or if they do DI at all.

Lukasz and I have been working together recently and he asked for my views on this thread. I totally agree that 1:1 mapping between interfaces and classes is pointless and wasteful. Where I do find interfaces valuable is ensuring that `a-stub-implementation` matches `a-real-implementation`, and then having `a-contract-test` that makes sure that I can trust my stub.

In a strongly typed language like Java you only have interfaces and polymorphism to allow you to do this. You don't really need interfaces in JS since the test should fail when there's a disparity between the stub and the real, but having TypeScript provide feedback while coding is definitely a plus and IMO is worth defining an interface for when you have multiple implementations.

PR #402, opened 2019-01-12
allow for deeply nested test double objects

lgandecki:
Popular Prisma [graphqlgen](https://github.com/prisma/graphqlgen) scaffolds a graphql server with a Context interface, that is used by graphql resolvers to execute commands and queries (on services, repositories, etc).

the default generated example looks like this:
```typescript
export interface Context {
  data: Data
}

export interface User {
  id: string
  name: string | null
  postIDs: string[]
}

export interface Post {
  id: string
  title: string
  content: string
  published: boolean
  authorId: string
}

export interface Data {
  posts: Post[]
  users: User[]
  idProvider: () => string
}

```
With this change I can do things like:

```typescript

const context = td.object<Context>();

context.data.users = [{id: "userId", name: "Hello", postIDs: []}]
td.when(context.data.idProvider()).thenReturn("providerId")
```

And everything is still properly typed. :-) 
Meaning - if I do 
```typescript
context.data.users = [{ids: "userId"}]
```

I get `TS2322: Type '{ ids: string; }' is not assignable to type 'User'.`

and for
```typescript
context.data.users = [{id: "userId", name: "Hello", postIDs: [1]}]
```
I get `TS2322: Type 'number' is not assignable to type 'string'.` error, with the `1` marked in the IDE as an error.

for 
```typescript
   td.when(context.data.idProvider("sdf")).thenReturn("providerId")
```
I get `TS2554: Expected 0 arguments, but got 1.`

...and so on.

This context type that  holds repositories/services is a common graphql pattern. I'm pretty sure there could be other uses as well.

### DIFF
--- a/src/object.js
+++ b/src/object.js
@@ -41,14 +41,12 @@ var createTestDoubleViaProxy = (name, config) => {
     if (!target.hasOwnProperty(propKey) && !_.includes(config.excludeMethods, propKey)) {
       let nameWithProp = `${nameOf(internalName)}.${String(propKey)}`
       const obj = tdFunction(nameWithProp)
-      obj.___mocks___ = {}
       target[propKey] = new Proxy(obj, generateHandler(nameWithProp))
     }
     return target[propKey]
   }
 
   const obj = {}
-  obj.___mocks___ = {}
   return new Proxy(obj, generateHandler(name))
 }
 
--- a/src/object.js
+++ b/src/object.js
@@ -32,15 +32,24 @@ var createTestDoublesForFunctionNames = (names) =>
 
 var createTestDoubleViaProxy = (name, config) => {
   ensureProxySupport(name)
-  const obj = {}
-  return new Proxy(obj, {
-    get (target, propKey, receiver) {
-      if (!obj.hasOwnProperty(propKey) && !_.includes(config.excludeMethods, propKey)) {
-        obj[propKey] = tdFunction(`${nameOf(name)}.${String(propKey)}`)
-      }
-      return obj[propKey]
-    }
+
+  const generateHandler = (internalName) => ({
+    get: (target, propKey) => generateGet(target, propKey, internalName)
   })
+
+  const generateGet = (target, propKey, internalName) => {
+    if (!target.hasOwnProperty(propKey) && !_.includes(config.excludeMethods, propKey)) {
+      let nameWithProp = `${nameOf(internalName)}.${String(propKey)}`
+      const obj = tdFunction(nameWithProp)
+      obj.___mocks___ = {}
+      target[propKey] = new Proxy(obj, generateHandler(nameWithProp))
+    }
+    return target[propKey]
+  }
+
+  const obj = {}
+  obj.___mocks___ = {}
+  return new Proxy(obj, generateHandler(name))
 }
 
 var ensureProxySupport = (name) => {
--- a/test/safe/object.test.js
+++ b/test/safe/object.test.js
@@ -124,9 +124,20 @@ module.exports = {
       td.when(testDouble.something()).thenReturn('Original Something')
       td.when(testDouble.different.something()).thenReturn('Different Something')
 
-      testDouble.___mocks___ = {}
       assert._isEqual(testDouble.something(), 'Original Something')
       assert._isEqual(testDouble.different.something(), 'Different Something')
+    },
+    'Resets all deeply nested test doubles with td.reset' () {
+      if (!global.Proxy) return
+      testDouble = td.object()
+
+      td.when(testDouble.something()).thenReturn('Original Something')
+      td.when(testDouble.different.something()).thenReturn('Different Something')
+
+      td.reset()
+
+      assert._isEqual(testDouble.something(), undefined)
+      assert._isEqual(testDouble.different.something(), undefined)
     }
   }
 }
--- a/test/safe/object.test.js
+++ b/test/safe/object.test.js
@@ -107,6 +107,26 @@ module.exports = {
       } catch (e) {
         assert._isEqual(e.message, "Error: testdouble.js - td.object - The current runtime does not have Proxy support, which is what\ntestdouble.js depends on when a string name is passed to `td.object()`.\n\nMore details here:\n  https://github.com/testdouble/testdouble.js/blob/master/docs/4-creating-test-doubles.md#objectobjectname\n\nDid you mean `td.object(['Woah'])`?")
       }
+    },
+    'Allow for deeply nested test double objects' () {
+      if (!global.Proxy) return
+      testDouble = td.object()
+
+      td.when(testDouble.something.very.deeply.nested()).thenReturn('nay!')
+
+      assert._isEqual(testDouble.something.very.deeply.nested(), 'nay!')
+      assert._isEqual(testDouble.something.very.deeply.nested.toString(), '[test double for ".something.very.deeply.nested"]')
+    },
+    'Deeply nested test double objects work also when its property names are repeated' () {
+      if (!global.Proxy) return
+      testDouble = td.object()
+
+      td.when(testDouble.something()).thenReturn('Original Something')
+      td.when(testDouble.different.something()).thenReturn('Different Something')
+
+      testDouble.___mocks___ = {}
+      assert._isEqual(testDouble.something(), 'Original Something')
+      assert._isEqual(testDouble.different.something(), 'Different Something')
     }
   }
 }


searls:
Thanks for this. I haven't touched this part of the code in a long time, so I'll need to be able to set aside some time to have a think about this implementation. 

The thing about it is that `td.imitate`-based fakes do what you're describing, but based on modeling itself after a real thing, not leaning on the compiler type info. So I do understand the value here—mostly wondering what impact this might have on non-typed uses of the `Proxy` API that might be out there

lgandecki:
Sure, no problem. I was thinking about possible backward compability, I think the tests are covering all the possible cases. unless someone was using the fact, that there would be a exception while trying to do 
const something = td.object();
something.something.something();

But I doubt it. 

Take your time and if you think about some edge case I haven't covered, please do share. For the project I'm working on currently this is fantastically convenient way of mocking. :) So I will probably use a fork for now..

searls:
FWIW, there are a lot of people who think "mocks that return mocks" is a pretty major antipattern/test smell. I don't really care about the issue, certainly not enough to think td.js should break for that use case by default

lgandecki:
I don't think this is a mock that returns a mock though, is it?
 It's just a test double of a certain shape, on runtime simulating almost "any" possible shape, but during the "compilation" time it's shape has to conform to a specific type, even if its a collection of types.
I'm not doing (nor recommending)  td.when().thenReturn(td.func())

src/object.js
```diff
@@ -32,15 +32,24 @@ var createTestDoublesForFunctionNames = (names) =>
...
+
+  const generateHandler = (internalName) => ({
+    get: (target, propKey) => generateGet(target, propKey, internalName)
   })
+
+  const generateGet = (target, propKey, internalName) => {
+    if (!target.hasOwnProperty(propKey) && !_.includes(config.excludeMethods, propKey)) {
+      let nameWithProp = `${nameOf(internalName)}.${String(propKey)}`
+      const obj = tdFunction(nameWithProp)
+      obj.___mocks___ = {}
```
searls: I'm just now getting around to reviewing this, @lgandecki. Sorry for being slow. The only red flag I see is this `___mocks___` scratchpad. So far the library has taken pretty extreme pains to avoid keeping track of _anything_ on the fakes it creates, instead storing them in memory in a private way that isn't exposed to the user. 

I like this feature, but I wouldn't want to break from that tradition here if we can avoid it. I don't have an immediate solution, but if you check any code described as a "store" elsewhere that's how we've been doing. Most important thing is that whatever it is be registered with `td.reset` to reset appropriately between tests.

lgandecki: Good Point! There was actually no need for this scratchpad anymore, it was a leftover of a previous implementation.. sorry! I just removed it.

I also added a test that verifies that td.reset does work properly here.

Please let me know if there is something more you would like to have changed here.

lgandecki:
Hello @searls do you have any more comments? I don't want to be pushy.. We are starting a new graphql-based project where we will need this functionality, so far I've been linking a local testdouble with the tweak but that's not a solution for once we get 4-6 people working on the project next week. 
I can easily publish my own npm version, but just thought I'd ask first, since you said you only had one red flag, that was easy to remove. :)

searls:
I'm sorry I haven't gotten to this yet! It's a big enough code change that I want to pull it down and use it locally before merging/releasing, but I haven't had the time yet. I hope I get the chance tomorrow. Feel free to keep bugging me if I'm not responsive

searls:
Howdy -- spent a couple hours working on this this morning and I've found an issue, namely that the proxy now returns undefined for a given function property.

For instance, this is the behavior of tdjs on master (running `script/repl`):

```js
💚  Let's play with testdouble.js!  💚
td > td.explain(td.object('lol').foo)
{ name: 'lol.foo',
  callCount: 0,
  calls: [],
  description:
   'This test double `lol.foo` has 0 stubbings and 0 invocations.',
  children:
   { toString:
      { name: undefined,
        callCount: 0,
        calls: [],
        description: 'This is not a test double function.',
        isTestDouble: false } },
  isTestDouble: true }
```

```js
td > td.explain(td.object('lol').foo)
{ name: undefined,
  callCount: 0,
  calls: [],
  description: 'This is not a test double function.',
  isTestDouble: false }
> td.object('lol').foo
Proxy [ { [Function: testDouble] toString: [Function] },
  { get: [Function: get] } ]
```

The reason for this is that `td.explain()` is looking in `src/store`'s `for()` function for the test double function and because it is wrapped by a proxy, none is found.

searls:
I've pushed up a failing test in 3bbed08fb3f806c1864e70ba99ca545ed3f6b8ee if you want to look at it

searls:
Another wrinkle here is that the blank slateness of `Proxy` makes debugging pretty hard.

```
    console.log('a', context)
    console.log('b', context.data)
    console.log('c', context.data.idProvider)
```

Will each print `undefined` but `=== undefined` will return false. Maybe we should add a default `toString()` impl?

searls:
Blah, this is not easy. Even if we patch the store such that aliases can be stored (like this patch does):

```
diff --git a/src/object/proxy.ts b/src/object/proxy.ts
index 972ab4d..f56d193 100644
--- a/src/object/proxy.ts
+++ b/src/object/proxy.ts
@@ -2,6 +2,7 @@ import * as theredoc from 'theredoc'
 import _ from '../wrap/lodash'
 import log from '../log'
 import tdFunction from '../function'
+import store from '../store'
 
 export default function proxy<T> (name: string, { excludeMethods } : { excludeMethods?: string[] } = {}) : T {
   ensureProxySupport(name)
@@ -31,7 +32,10 @@ const generateHandler = (internalName, excludeMethods) => ({
 const generateGet = (target, propKey, internalName, excludeMethods) => {
   if (!target.hasOwnProperty(propKey) && !_.includes(excludeMethods, propKey)) {
     const nameWithProp = `${internalName || ''}.${String(propKey)}`
-    target[propKey] = new Proxy(tdFunction(nameWithProp), generateHandler(nameWithProp, excludeMethods))
+    const tdFunc = tdFunction(nameWithProp)
+    const tdFuncProxy = new Proxy(tdFunc, generateHandler(nameWithProp, excludeMethods))
+    store.registerAlias(tdFunc, tdFuncProxy)
+    target[propKey] = tdFuncProxy
   }
   return target[propKey]
 }
diff --git a/src/store/index.js b/src/store/index.js
index 2818520..d8ba107 100644
--- a/src/store/index.js
+++ b/src/store/index.js
@@ -4,7 +4,7 @@ import { EventEmitter } from 'events'
 const storeEmitter = new EventEmitter()
 let globalStore = []
 
-export default {
+const store = {
   onReset (func) {
     storeEmitter.on('reset', func)
   },
@@ -15,7 +15,7 @@ export default {
   },
 
   for (testDouble, createIfNew = true) {
-    const entry = _.find(globalStore, { testDouble })
+    const entry = _.find(globalStore, e => testDouble === e.testDouble || testDouble === e.alias)
     if (entry) {
       return entry
     } else if (createIfNew) {
@@ -28,5 +28,11 @@ export default {
         return globalStore.push(newEntry)
       })
     }
+  },
+
+  registerAlias (testDouble, alias) {
+    store.for(testDouble).alias = alias
   }
 }
+
+export default store
```

Then we still just run into a call stack exceeded when traversing the children of the proxy in explain itself (oddly enough this feature was added yesterday):

```js
function explainChildren (thing) {
  const explanations = []
  const children = _.cloneDeepWith(thing, (val, key, obj, stack) => {
    if (_.isFunction(val) && stack) {
      return _.tap(explainFunction(val), (explanation) => {
        if (explanation.isTestDouble) explanations.push(explanation)
      })
    }
  })
```

lgandecki:
Hey Justin, thank you so much for working on this!

I got all the tests to pass with a few changes, which you will probably flag again, but maybe it's a start?

https://github.com/testdouble/testdouble.js/pull/406

Looks like the cloneDeep from lodash doesn't play nicely with proxies.

I don't know yet what to do about the fact that console.log shows undefined. I tried to play around with it but have to stop. I will probably have some time to dig into this again tomorrow.

searls:
To simplify working together, I've added you as a collaborator on the repo with commit bit (🎉!), and opened a PR against the branch I was working on. Feel free to just continue work there directly so we can tighten the feedback loop.

Closing in favor of #407

lgandecki:
Thanks so much @searls :-) I pushed my changes to #407 , I hope you think that's an acceptable direction.

Issue #367, opened 2018-05-06
Remove es6-map polyfill


searls:
It looks like es6-map is now supported sufficiently on all evergreen browsers and supported Node versions, so we ought to be able to safely remove our es6-map dependency

http://kangax.github.io/compat-table/es6/#test-Map

primarilysnark:
This is now resolved by #405.

searls:
You probably know this, but an FYI in case not -- if you had written "Fixes #367" (or "Resolves") into the body of the PR it'd have auto closed this issue

PR #405, opened 2019-01-29
Remove unnecessary dependency on es6-map.

primarilysnark:
As a follow-up to #367, this PR removes the `es6-map` dependency from `testdouble.js`.


searls:
thanks!

PR #417, opened 2019-06-08
add cloneArgs option to td.when/td.verify

### LINKED ISSUE #375: Arguments of mocked functions are saved  by reference.
### Description
Arguments of mocked functions are saved  by reference. 

```
const f = td.function();
p.some = "1";
f(p)
p.some = "2" 
f(p)
```

I expect `explain` to show calls to `f` with `some="1"` and `some="2"`. However, it shows it with `some="2"` twice.  

### Issue

The code example below printed the following:
name: Salvador last: Sobral  <-- first call
name: Netta last: Barzilai
Call 1. name: Netta last: Barzilai  <-- first call according to expect
Call 2. name: Netta last: Barzilai

### Environment

- [x] `node -v` output: v8.11.2
- [x] `npm -v` (or `yarn --version`) output: 5.6.0
- [x] `npm ls testdouble` (or `yarn list testdouble`) version: testdouble@3.8.1 

#### Code-fenced Examples

```js
var td = require('testdouble')

class Person { }
const p = new Person();

const f = td.function();
td.when(f(td.matchers.anything())).thenDo((p) => { console.log("name: " + p.first + " last: " + p.last); });

// first call
p.first = "Salvador";
p.last = "Sobral";
f(p);

// second call 
p.first = "Netta";
p.last = "Barzilai";
f(p);

// explain has "Netta Barzilai" twice
td
  .explain(f)
  .calls
  .forEach((call, index) => {
  call.args.map((arg) => console.log("Call " + (index + 1) + ". name: " + arg.first + " last: " + arg.last));
});
```

### LINKED ISSUE #416: Challenges testing mutable objects
### Description
When testing functions that modify input objects, I don't see a way to validate my code without rewriting due to the fact that testdoubles store arguments by reference instead of creating a copy. Largely a duplicate of #375 , but since that focused on td.explain and I can't re-open, creating as a new issue.  Basically, I just spent a couple of hours trying to figure out why I couldn't get a test to pass because of this and it made me salty.

### Issue

Here's a simple test case that is similar, conceptually, to the function I was trying to test, and hopefully not a crazy enough use case that this is easy to dismiss. My imaginary use case is part of a system where we try to acquire a lock on an object before operating on it; we have many types of objects that can be locked just by setting "status = locked" on the object and saving to the database, so we want to create a single function to handle the locking behavior:

```
function lockAndDo ( dataObject, saveFunction, doFunction ) {
  dataObject.status = "LOCKED"
  saveFunction( dataObject )
  doFunction( dataObject )
  dataObject.status = "UNLOCKED"
  saveFunction( dataObject )
}
```

Obviously, in real life, this method would have a lot more error-checking and so forth.  So, here's my failing test:

```
module.exports = {
  LockingWorkflow: () => {
    const data = { status: "UNLOCKED" }
    const saveFunc = td.func()
    const doFunc = td.func()

    lockAndDo( data, saveFunc, doFunc )

    td.verify( saveFunc( td.matchers.contains( { status: "LOCKED" } ) ) )
    td.verify( doFunc( data ) )
    td.verify( saveFunc( td.matchers.contains( { status: "UNLOCKED" } ) ) )
  }
}
```

Due to the behavior described in #375 , this test fails on the first `td.verify` call because the second invocation changes the status on `dataObject` back to UNLOCKED.

Unless I'm missing something, if I'm trying to write a function that has to work this way (maybe I'm using somebody else's data library), my *production* code has to make a copy of `dataObject` between locking and unlocking the object in order to have a test to validate this sequence. I'd much rather take the performance hit of having the test library clone objects!

searls:
Fixes #375 
Fixes #416

Note that this will be a performance regression for anybody calling test doubles with large or hard to deep-clone arguments. Since that falls firmly in the category of "not how td.js is designed to be used", I'm going to put this out there and then see what hell it wreaks rather than aggressively benchmark now.

### DIFF
--- a/src/store/calls.js
+++ b/src/store/calls.js
@@ -7,7 +7,7 @@ store.onReset(() => { callHistory = [] })
 
 export default {
   log (testDouble, args, context) {
-    store.for(testDouble).calls.push({ args, context })
+    store.for(testDouble).calls.push({ args, context, cloneArgs: _.cloneDeep(args) })
     return callHistory.push({ testDouble, args, context })
   },
 
@@ -30,7 +30,8 @@ export default {
 
   where (testDouble, args, config) {
     return _.filter(store.for(testDouble).calls, function (call) {
-      return argsMatch(args, call.args, config)
+      const pastArgs = config.cloneArgs ? call.cloneArgs : call.args
+      return argsMatch(args, pastArgs, config)
     })
   },
 
--- a/docs/5-stubbing-results.md
+++ b/docs/5-stubbing-results.md
@@ -615,6 +615,45 @@ fetch('/B', function (er, result) {}) // will be invoked 2nd
 fetch('/C').then(function (result) {}) // will be invoked 1st
 ```
 
+#### cloneArgs
+
+Every now and then, the code under test will mutate the object you initially
+pass into a stubbing configuration, and you want to be sure that when the test
+double function is invoked by your code, the stubbing is determined to be
+satisfied (or not) by comparing the actual call against the value at the time
+you configured it.
+
+Confused yet? Here's a quick example of how testdouble.js behaves by default:
+
+```js
+const func = td.func()
+const person = { age: 17 }
+td.when(func(person)).thenReturn('minor')
+
+// Later, in your code
+person.age = 30
+func(person) // 'minor'
+```
+
+Maybe you don't want this! Maybe you want to be sure the stubbing is only
+satisfied so long as the arguments are exactly as they were when you configured
+the stubbing. You can do that! By setting `cloneArgs` to true, you can do the
+following:
+
+```js
+const func = td.func()
+const person = { age: 17 }
+td.when(func(person), { cloneArgs: true }).thenReturn('minor')
+
+// Later, in your code
+person.age = 30
+func(person) // undefined
+```
+
+While passing in data with an expectation that it be mutated by your subject is
+generally Not Recommended™, this option should enable edge cases where mutation
+is unavoidable or out of your control.
+
 ## Congratulations!
 
 And that's about all there is to say about stubbing. Great news, because the
--- a/docs/6-verifying-invocations.md
+++ b/docs/6-verifying-invocations.md
@@ -368,6 +368,41 @@ var doNotCall = td.function()
 td.verify(doNotCall(), {times: 0, ignoreExtraArgs: true}) // passes
 ```
 
+
+#### cloneArgs
+
+What if you want to verify a call took place and the subject (for better or
+worse) mutated an argument after it was passed to the test double function?
+Since testdouble.js saves arguments by reference by default, you won't get the
+result you want:
+
+```js
+const func = td.func()
+const person = { age: 17 }
+
+// later, in your code
+func(person)
+person.age = 30
+
+// back in your test
+td.verify(func({ age: 17 })) // 💥 Test failure! td.js recorded age as 30!
+```
+
+For cases like these, you can work around the mutation by setting `cloneArgs` to
+`true`:
+
+```js
+const func = td.func()
+const person = { age: 17 }
+
+// later, in your code
+func(person)
+person.age = 30
+
+// back in your test
+td.verify(func({ age: 17 }), { cloneArgs: true }) // 😌 all good
+```
+
 ## Congratulations!
 
 And that's everything there is to know about verifying behavior with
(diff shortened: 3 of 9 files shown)


monovertex:
Hello, @searls! I know this is an old update, but we've just upgraded from 3.8.1 to the latest version and this is breaking for part of our testing suite. We're using testdouble in an Ember app, where sometimes the arguments to our stubs are Ember objects. 

These have a chain of references to the entire app registry and even the module registry, which cloneDeep then tries to clone and and some point encounters an undefined reference, throwing an error. Regardless of that specific error, cloning the entire context of the app is definitely bad. 

Is there any way to avoid calling `cloneDeep` when the `cloneArgs` config option is not `true`? We were thinking either somehow accessing the `config` object in the `log` function, or perhaps having a global config (if possible) to turn off the cloning feature altogether?

We can definitely open a PR in this direction, we just need a bit of guidance on how would this be implemented. I'm guessing it's not trivial, considering that this concern is expressed in the PR description.

searls:
Hey @monovertex sorry to hear that you're having this problem. It has literally been two years since I've even thought about this aspect of the library so I need some help understanding your need. Could you provide a few things?

1. An example test that's not working as you expect
2. Any workaround you've found so far
3. A proposed API change

I got confused because on its face: "Is there any way to avoid calling cloneDeep when the cloneArgs config option is not true" sure seems like that's how it should behave, so I'm afraid I don't understand

src/store/calls.js
```diff
@@ -7,7 +7,7 @@ store.onReset(() => { callHistory = [] })
 
 export default {
   log (testDouble, args, context) {
-    store.for(testDouble).calls.push({ args, context })
+    store.for(testDouble).calls.push({ args, context, cloneArgs: _.cloneDeep(args) })
```
monovertex, line 10: Thank you for your prompt reply @searls! The current implementation will always call `cloneDeep`, even when `config.cloneArgs` is `false`. True, it will only use the cloned args when that option is set, but the cloning of the arguments happens all the time. This is also highlighted by the original concern you expressed about this change having a performance impact for calling test doubles with large or hard to deep-clone arguments.

Unfortunately the current workaround was to downgrade back to 3.8.1, as there's no way to control the deep cloning.

I am not necessarily proposing an API change, but rather to see if the cloning could only be done when `cloneArgs` is `true`. From the code I referenced here, it seems that the `config` object is not available, so perhaps there's an architectural limitation?

[Here is a very short reproduction](https://jsfiddle.net/ocb1qagy/2/) that evidences how the cloning might impact a test. Obviously it's a ridiculous example, but in our particular case, there's a chain of references that leads to a similar prop definition, where `cloneDeep` then promptly throws an error.

Now, we would never be able to use the `cloneArgs` config option, that is true, but that is okay, we never needed it anyway, and we feel like the corect handling in those situation would be instead to use `td.matchers.argThat` and assert on particular properties we want to stay the same.

searls: Ahhhhhhhh, I get it. 

Yeah this is an ancillary benefit of these changes that's not directly related to the functionality. The goal here that we were solving for was error messages and `td.explain` would lie to you if something passed to a test double was mutated during the test before the printout. Cloning the arguments is a way to ensure that the library's messages are accurate, which is indeed very important (as having test output that says stuff like "expected foo({age: 32}) but was actually foo({age: 32})" can do a lot of arm when debugging)

Yeah, I don't have a great solution here. If the issue is the error, then I think we can wrap cloneDeep in a try-catch and fall back on the original args if cloneArgs is some symbol that represents a failed processing.

monovertex: That doesn't sound like a bad idea actually. Although in our case we would probably prefer to maybe have a feature flag to turn off the deep cloning altogether, as it's something that has potential to slow down the app and maybe even cause  huge memory leaks if a reference to the cloned object is retained somewhere, if the cloning is successful.

In our particular case, where we can't control the objects and their references (they're generated by Ember), we prefer the tradeoff of potentially inaccurate messages instead of the performance impact.

searls: Yeah, I understand the performance concern but in general this library really tries to reduce or eliminate global configuration settings wherever possible so I'm not comfortable adding a setting that could decrease the veracity of its messages

searls: If you're willing to wrap this in error handling and add a test I'd be happy to pick up a PR and take it from there, though

monovertex: Alright, I will add the try/catch, and we'll see how the test suite performs after that. If the performance hit is major, we'll most likely fork the library and add our own config guard.

monovertex: @searls I opened #469, we can continue the discussion there. Thanks!

Issue #370, opened 2018-05-16
td.object(['one', 'two']) syntax creates array-like object


Schoonology:
Surprisingly, `td.object({ something() {} })` and `td.object(['something'])` produce different results; the latter is an Array.

```
> td.object({ something() {} }).splice
undefined
> td.object(['something']).splice
[Function: splice]
```

searls:
Huh. That seems pretty wrong. Looks like it's a side effect of the (out-of-context, odd) decision to use `_.transform` to create the return object when passed an array of names:

https://github.com/testdouble/testdouble.js/blob/master/src/object.js#L29-L31

giltayar:
Stale. Closing. Please reopen if still relevant and I will look into it.

PR #421, opened 2019-07-03
[Issue-370] - Update createTestDoublesForFunctionNames to return an object

bannmoore:
cc @primarilysnark 

This PR addresses [Issue 370](https://github.com/testdouble/testdouble.js/issues/370).

Context: When `td.object` is called with an array of function names, it returns an array; my understanding of the issue is that `td.object` should always return an object.

From the original issue:

```
> td.object({ something() {} }).splice
undefined
> td.object(['something']).splice
[Function: splice]
```

Included Changes:
- Add an assertion to the `td.object` w/ array of function names test, to confirm that the resulting object is _not_ an array.
- Add a base value of `{}` to the accumulator in `createTestDoublesForFunctionNames`, to ensure that it responds with an object.


searls:
Landed in v3.12.1

PR #422, opened 2019-07-15
Run CI also on Node 10

lumaxis:
Given that Node 10 is the current LTS, it's probably a good idea to also test against it.


lumaxis:
Oh, I guess this is using Circle CI now 🤔

Updating …

searls:
yes feel free to nuke the old travis yml wups

PR #427, opened 2019-08-05
Fix .verify() and .when() when using testdouble objects as arguments

### LINKED ISSUE #419: td.object() cannot be used as an argument inside td.when(...)
### Description

I am trying to mock the return value of a function that accepts an argument.

### Issue

I cannot mock the return value of a function with an argument when using td.object() as the argument.

E.g.
```js
arg = td.object();
td.when(mocked(arg)).thenReturn('mocked');
mocked(arg) --> returns undefined
```

The issue only appears when `arg = td.object()`. The code is successfully executed when using a generic object like `arg = {}` or `arg = "some-string"`

### Environment

- [ ] `node -v` output: v8.15.0
- [ ] `npm -v` (or `yarn --version`) output: 6.9.0
- [ ] `npm ls testdouble` (or `yarn list testdouble`) version: testdouble@3.12.0

#### Code-fenced Examples

The following example showcases the issue:

```js
const td = require('testdouble');

function check(desc, arg) {
  const mocked = td.function();
  td.when(mocked(arg)).thenReturn('mocked');
  console.log(desc, '::', mocked(arg), ':: match:', mocked(arg) === 'mocked');
}

check('{}', {});
check('"string"', 'string');
check('td.object()', td.object());
```

The output of executing the example:
```
{} :: mocked :: match: true
"string" :: mocked :: match: true
td.object() :: undefined :: match: false
```

### LINKED ISSUE #426: Failing verification when using td.object() as argument
### Description

When passing a `td.object()` as a parameter to a function that then call a callback with that function, verifying the invocation on that callback fails.
I attached an example that repros the issue in the repl and is fairly similar to my real world case. It's very possible that this is also just me misusing the library but I imagine it's fairly easy to fall into this trap? 🤔

Additionally, this is possibly a duplicate of #419 but I decided to open a separate issue instead since this is when using `td.verify()`, not `td.when()`.

### Issue

Testdouble fails to correctly verify the invocation even though the function was seemingly correctly called. Additionally, the `(undefined)` name in the error message is confusing as it makes it very hard to understand what exactly the problem is/in what way the arguments don't match.

```
Error: Unsatisfied verification on test double `logger`.

  Wanted:
    - called with `(undefined)`.

  All calls of the test double, in order were:
    - called with `(undefined)`.
    at Object.exports.default [as verify] (/Users/lukas/code/testdouble.js/src/verify.js:17:9)
    at Object.fail (/Users/lukas/code/testdouble.js/src/log.js:17:11)
```

### Environment

- [x] `node -v` output: `v10.11.0`
- [x] `npm -v` (or `yarn --version`) output: `6.4.1`
- [x] `npm ls testdouble` (or `yarn list testdouble`) version: `3.12.3`

#### Code-fenced Examples

```js
function middlewareFactory(logger) {
  return (res) => {
    if (logger) { logger(res); }
  };
}

const fakeResponse = td.object('response');
const fakeLogger = td.function('logger');

const middleware = middlewareFactory(fakeLogger);
middleware(fakeResponse);

td.verify(fakeLogger(fakeResponse));
```

lumaxis:
Fixes #419 and fixes #426

### DIFF
--- a/src/matchers/is-matcher.js
+++ b/src/matchers/is-matcher.js
@@ -1,2 +1,2 @@
 export default (thing) =>
-  thing && thing.__matches
+  thing && !thing[Symbol("__is_proxy")] && thing.__matches
--- a/src/matchers/is-matcher.js
+++ b/src/matchers/is-matcher.js
@@ -1,2 +1,2 @@
 export default (thing) =>
-  thing && !thing[Symbol("__is_proxy")] && thing.__matches
+  thing && !thing[Symbol('__is_proxy')] && thing.__matches
--- a/src/object/proxy.ts
+++ b/src/object/proxy.ts
@@ -30,8 +30,8 @@ const generateHandler = (internalName, excludeMethods) => ({
 })
 
 const generateGet = (target, propKey, internalName, excludeMethods) => {
-  if (propKey === Symbol("__is_proxy")) {
-    return true;
+  if (propKey === Symbol('__is_proxy')) {
+    return true
   }
 
   if (!Object.prototype.hasOwnProperty.call(target, propKey) &&
--- a/src/object/proxy.ts
+++ b/src/object/proxy.ts
@@ -30,6 +30,10 @@ const generateHandler = (internalName, excludeMethods) => ({
 })
 
 const generateGet = (target, propKey, internalName, excludeMethods) => {
+  if (propKey === Symbol("__is_proxy")) {
+    return true;
+  }
+
   if (!Object.prototype.hasOwnProperty.call(target, propKey) &&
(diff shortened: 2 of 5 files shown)


lumaxis:
@searls Anything I can do to help move this forward? Would be great to finally get it merged 😊

searls:
Whoops! I somehow just never saw this. Short, sweet, fixes build, fixes two issues, what's not to like

[Prediction: we'll find out that referencing `Symbol` is going to break somebody somewhere, ¯\_(ツ)_/¯ ]

searls:
Landed in 3.13.1

lumaxis:
Hah, thanks 🙂 Fingers crossed!

PR #436, opened 2019-12-30
Missing TypeScript declaration for td.matchers.create() added

akloeber:
TypeScript declaration for `td.matchers.create()` has been added according to the API described at https://github.com/testdouble/testdouble.js/blob/master/docs/8-custom-matchers.md#tdmatcherscreate-api.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,13 +32,21 @@ export interface Explanation {
   isTestDouble: boolean;
 }
 
+export interface MatcherConfig {
+  matches(matcherArgs: any[], actual: any): boolean;
+  name?: string | ((matcherArgs: any[]) => string);
+  onCreate?(matcherInstance: any, matcherArgs: any[]): void;
+  afterSatisfaction?(matcherArgs: any[], actual: any): void;
+}
+
 export interface Matchers {
   anything(): any;
   isA(type: Function): any;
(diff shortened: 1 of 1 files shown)


searls:
Hey, @lgandecki can you spot check this for us?

lgandecki:
Sure thing. It looks correct, I also gave this definition a try with the built-in matchers: https://github.com/testdouble/testdouble.js/tree/master/src/matchers/builtin and there were no type errors.

This could be better - it would be nice to type `matcherInstance: any, matcherArgs: any[]` instead of setting them to any, but it's definitely a step into the right direction and will at least get rid of type errors.

searls:
did you mean to close this?

akloeber:
@lgandecki I typed `matcherInstance` as `any` as yet there is no type for matcher instances defined. The build-in matcher-factories defined in the `Matchers`-interface have return-type `any` too. Here a dedicated `Matcher`-type would make sense but I didn't know about the public shape / API of these. Maybe there should be symbolic type alias (e.g. `type Matcher = any;`) to make the type definitions more self-explanatory.

Regarding `matcherArgs: any[]`: I don't see a better way in typing this as this could really be anything (no arg at all, or extensive list of args with varying types).

lgandecki:
@searls no, sorry, good to merge!

Issue #447, opened 2020-05-25
Add td.instance(constructorFtn)?


woldie:
### Description
Would like a one liner for when I don't need the mock constructor and just want a mock instance of a real constructor.

### Issue
What I do today:
```
let MockMarsRover = td.constructor(MarsRover);
let marsRover = new MockMarsRover();
```
where `MarsRover` is a constructor function.

What I'd _also_ like to be able to do:
```
let marsRover = td.instance(MarsRover);
```

@searls your "Please don't mock me" presentation was great!  725 slides, wew!!  Outside-In TDD is my jam.  I shared that prezzo with my agile coach friend and he switched from Sinon to testdouble.js immediately.

searls:
Thanks for the kind words.

I think I'd be ok with a PR that implemented `td.instance` as described if it conformed to the style of the repo and had tests

woldie:
Is this the correct way to run tests?

`npm run compile && npm run test`

woldie:
twas merged

searls:
Thanks for tidying this!

woldie:
Sure, I wrote you another one though, heheh.  https://github.com/testdouble/testdouble.js/issues/467

PR #448, opened 2020-05-25
Add td.instance()

woldie:
Attempt to implement #447

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ signup = td.replace(app, 'signup', {
 })
 ```
 
-### `td.func()`, `td.object()`, `td.constructor()`, and `td.imitate()` to create test doubles
+### `td.func()`, `td.object()`, `td.constructor()`, `td.instance()` and `td.imitate()` to create test doubles
 
 `td.replace()`'s imitation and injection convenience is great when your
 project's build configuration allows for it, but in many cases you'll want or
@@ -330,6 +330,28 @@ const subject = function (SomeConstructor) {
   const thing = new SomeConstructor()
   return thing.doStuff() // returns "ok"
 }
+```
+
+#### `td.instance()`
+
+If your code depends on ES classes or functions, then the `td.instance()` function 
+will create a mock constructor and return a new instance of that mock constructor.
+
+The following code snippets are functionally equiavalent:
+
+```js
+const fakeObject = td.instance(RealConstructor)
+td.when(fakeObject.doStuff()).thenReturn('just did it')
+
+fakeObject.doStuff() // returns "just did it"
+```
+
+```js
+const FakeConstructor = td.constructor(RealConstructor)
+const fakeObject = new FakeConstructor()
+td.when(fakeObject.doStuff()).thenReturn('just did it')
+
+fakeObject.doStuff() // returns "just did it"
 ```
 
 #### `td.imitate()`
--- a/index.d.ts
+++ b/index.d.ts
@@ -136,6 +136,23 @@ export function constructor<T>(
 ): TestDoubleConstructor<T>;
 
 //
+// fake: instance objects
+//
+
+/**
+ * Construct an instance of a faked class.
+ *
+ * @export
+ * @template T
+ * @param {{ new (...args: any[]): T }} constructor
+ * @returns {DoubledObject<typeof T>}
+ */
+export function instance<T>(
+  constructor: Constructor<T>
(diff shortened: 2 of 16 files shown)


searls:
Thanks @woldie -- could you take a look at the build and see what's failing? `npm run test:ci` locally is the easiest command to imitate it

woldie:
Yeah, sorry about that, I thought I had all tests passing.  Ok, attempt 2 coming soon.

README.md
```diff
@@ -332,6 +332,28 @@ const subject = function (SomeConstructor) {
 }
 ```
 
+#### `td.instance()`
```
woldie, line 335: Please help me refine this documentation section

index.d.ts
```diff
@@ -135,6 +135,23 @@ export function constructor<T>(
...
+// fake: instance objects
+//
+
+/**
+ * Construct an instance of a faked class.
+ *
+ * @export
+ * @template T
+ * @param {{ new (...args: any[]): T }} constructor
+ * @returns {DoubledObject<typeof T>}
```
woldie, line 148: Was not sure what the return type should be here

test/safe/instance.test.js
```diff
@@ -0,0 +1,55 @@
+let Thing, SuperThing, fakeInstance
```
woldie, line 1: Pared down version of `constructor.test.js` with assertions I felt were safe to make

src/instance.js
```diff
@@ -0,0 +1,5 @@
+import constructor from './constructor.js'
+
+export default function instance (typeOrNames) {
+  return new (constructor(typeOrNames))()
+}
```
woldie, line 5: I can redo this if you'd like to recommend a style that better matches the rest of the codebase.  Can do the _.tap thing if you think it looks better.

index.d.ts
searls: @lgandecki as resident typescript typings maintainer could you please help with this / add a suggestion for what it should be?

lgandecki: Sure thing. Looks correct. I've also verified it by cloning the branch and playing around with it. 
Thanks for your work @woldie .
I guess you could possibly get similar results using td.object(Class), but that uses proxy.. and is a bit less safe if used without the additional help of typescript

lgandecki:
@woldie please also make sure to remove all the console.logs from your branch :)

index.d.ts
woldie: Thanks @lgandecki!  I agree it would be more convenient to have `td.object(Class)` just do it all, but I think it's probably important to cordon off this particular convenience method since it is only a wrapper for a call `td.constructor`, but then drops the constructor on the floor when it is done with it, intentionally hiding the ability to mock static class methods.

Looking at this particular change again, I think my use of `typeof T` is in the wrong spot for what I want here.  I use JSDoc to specify types in my home project and I cribbed from it, but I think I cribbed without checking my work.  `typeof T` describes a constructor of type `T` when used in conjunction with `T` by itself which is meant to indicate an instance of `T`.  When I do a good job with my JSDoc, it works just as well as TypeScript for most things in terms of the quality of IntelliSense and inspections within IntelliJ.

In my home project, I have the need to annotate a constructor, and I needed IntelliJ to understand that the return value from the annotation function was the constructor.  Here's how I expressed that in JSDoc:

```
/**
 * @param {typeof T} ctor constructor function.
 * @param {...(number|string)} args metadata about ctor.
 * @returns {typeof T} annotated constructor function
 * @template T
 */
Injector.prototype.annotateConstructor = function (ctor, args) {
  ...
}
```

If the JSDoc part of a TypeScript declaration understands `typeof T` the same way, then the TypeScript type declaration should read:

```
/**
 * Construct an instance of a faked class.
 *
 * @export
 * @template T
 * @param {typeof T} constructor
 * @returns {DoubledObject<T>}
 */
export function instance<T>(
  constructor: Constructor<T>
): DoubledObject<T>;
```

I'll try this out later today and see how IntelliJ interprets it.

lgandecki: Sounds good. I'm unfortunately not really familiar with JSDocs :-(

searls:
Landed in 3.16.0. Thanks @woldie!

woldie:
Just pulled 3.16.0 into my project, td.instance works great and many tests got simpler!  Thanks @searls and @lgandecki for the assist!

PR #449, opened 2020-07-30
docs: add `testdouble-qunit` to Plugins list

alexlafroscia:
Adds `testdouble-qunit` to list of plugins that integrate Testdouble with populate test runners


searls:
Thanks!

PR #452, opened 2020-08-31
fix NODE_OPTIONS example to specify the loader

luxferresum:
The example passed `node` to `NODE_OPTIONS` which is wrong and made it to not work.


searls:
Thanks!

PR #454, opened 2020-10-03
Easier on readers when following links

rhettc:
Heya, 
Newcomer to TD so please, don't mock me 😏
On my journey through the docs I landed [here](https://github.com/testdouble/testdouble.js/blob/master/docs/5-stubbing-results.md#loosening-stubbings-with-argument-matchers) coming from [here](https://github.com/testdouble/testdouble.js#tdwhen-for-stubbing-responses) and found myself confused because the first example above had nothing to do with it. Over the course of making the proposed edit I realized that the original text "examples above" is correct, but it's like three examples (code blocks) ago, and not any of the example invocations in the code block above.
I think this change could help, and even if the reader was coming from the top of the doc, they're just getting a little bit of reinforcement. And we can both agree that appropriately-spaced repetition is the best way to appease the Crabigator.

じゃあね

### DIFF
--- a/docs/5-stubbing-results.md
+++ b/docs/5-stubbing-results.md
@@ -95,9 +95,12 @@ As you can see, any additional invocations will continue to return the final stu
 
 #### A note on unconditional stubbing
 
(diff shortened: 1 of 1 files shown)


searls:
Sure, thanks. Happy testing!

PR #456, opened 2020-10-22
Fix Incorrect Formatting in Docs

conradbeach:
Because of the missing backtick, the wrong sections of text were being formatted as code.

**Before**
incorrectly_formatted_code

**After**
correctly_formatted_code


searls:
Thanks!!

PR #462, opened 2021-05-11
master -> main

rosston:
Update references to `master` to instead refer to `main`, since the default branch has now been renamed.

I'm opening this as a PR specifically because it changes actual code and want to make it visible. Given actual code changes, should this be released to npm too?


searls:
Thanks Ross! Since GitHub redirects the old URLs gracefully I don't think it's worth cutting a release over

Issue #441, opened 2020-03-30
How to setup a TypeScript test with Jest?


Zardddddd60:
### Description
I tried to test my app with Jest in Typescript, but failed to load typings of testdouble.
Since the typing file of testdouble is not included, then I manually add testdouble typings to TS context like this:

```json
{
    "compilerOptions": {
        "typeRoots": ["./node_modules/testdouble", "./node_modules/@types"],
        "types": ["jest", "node", "testdouble"]
    }
}
```
Jest works as expected, testdouble not.

Then I use `import td from 'testdouble`, but the injected method, `replace`, from `testdouble-jest`, is not available.

I also tried to add `td` to global context by defining a `global.d.ts` but failed.

I found `@types/Jest` declared some variables: 
```typescript
declare var beforeAll: jest.Lifecycle;
``` 
Then i could use `beforeAll`.

if needed, i can create a repo to reproduce this issue.

Is there an example or a documentation guiding me how could i test in TypeScript with Jest?

### Issue

need a documentation of how to setup a TS test with Jest.

Zardddddd60:
this seems more like a TS question, i solved this by adding a `global.d.ts` file(make sure file read by tsc):

```typescript
import * as testdouble from 'testdouble';

declare global {
    const td: typeof testdouble;
}

```
hope this helps others new to TS

searls:
Yay! My "wait 24 hours to read my e-mail closely" strategy pays off again!

Happy testing,

rene-leanix:
This doesn't seem to work for current TypeScript versions any more. The declaration seems to have changed and a "hack" is required to avoid the [TypeScript error on global augmentations](https://stackoverflow.com/q/57132428/6813271):

```ts
import type * as testdouble from 'testdouble';

export {};

declare global {
  namespace NodeJS {
    interface Global {
      td: typeof testdouble;
    }
  }
}
```

Adding `export as namespace testdouble;` to the testdouble index.d.ts will allow to simplify this significantly to:

```ts
/// <reference types="testdouble" />

declare namespace NodeJS {
  interface Global {
    td: typeof testdouble;
  }
}
```

I'll create a PR to this effect.

PR #464, opened 2021-06-01
Provide module namespace

rene-leanix:
Providing a namespace for the module as a whole allows to simplify defining a global instance of `td`. See [this comment](https://github.com/testdouble/testdouble.js/issues/441#issuecomment-851978147).


searls:
I'll take your word for it!

PR #466, opened 2021-06-02
docs: fix link to `testdouble-qunit` plugin

alexlafroscia:
A while back I simplified the `testdouble-qunit` repo to remove the monorepo structure (as the Ember-specific package that used to also live there became unnecessary). The original documentation link is no longer valid; this updates it to just point directly to the repo itself.


searls:
Thank you!

PR #469, opened 2021-08-23
Graciously fall back from failed cloneDeep

monovertex:
In some cases `cloneDeep` can fail, rendering the stub unusable. This PR adds error handling for that case, falling back to using the original `args` object by reference. See discussion in #417.

@searls, please let me know if there's other methods of error handling. I searched for `try/catch` blocks in the source and only found some that were not relevant for this particular use case.


monovertex:
> Released as 3.16.2

Thank you! Sorry for the late reply, was away on holidays.

PR #481, opened 2022-03-01
Add `thenReturn` to `PromiseStubber` type declaration

### LINKED ISSUE #478: Stubbing sequential promise results
## description

I want to suggest an enhancement about `thenReject` and `thenResolve` function. In the older version before 3.14. when I want to do a sequential promise return I can do `thenReturn(Promise.resolve('value1'), Promise.reject(Error('error1')))`. However, now it forced me to use `thenReject` and `thenResolve` for stubbing promise, I can't actually achieve it easily with these two functions. As the code shows below. it will always return the last stubbing promise result. I was expecting that it will be the same sequence as I stubbing them.

Right now I fix this problem by introducing a counter and `thenDo` function. Still, It's way messier than before. I hope it can be achieved by calling `thenReject` and `thenResolve` function in a more natural way.

## demo

env: 
- node version 12.22.5
- testdouble version 3.16.4

```typescript
const fetch = td.function();
td.when(fetch('test')).thenReject(Error('error1'));
td.when(fetch('test')).thenReject(Error('error2'));
Array.from(Array(2)).forEach((x, i) => {
    fetch('test').catch((error: Error)=>{
        console.debug(error.message);
    })
});
// console log I'm expecting error1 then error2
// error2
// error2
// I'm expecting error1 then error2

td.when(fetch('test')).thenResolve('return1');
td.when(fetch('test')).thenResolve('return2');
Array.from(Array(2)).forEach((x, i)=> {
    fetch('test').then((value: string) => {
        console.log(value);
    });
});
//console log
// return2
// return2
//I'm expecting return1 then return2

const fetch = td.function();
td.when(fetch('test')).thenResolve('return1');
td.when(fetch('test')).thenReject(Error('error1'));
Array.from(Array(2)).forEach((x, i) => {
    fetch('test')
    .then((value:string) => {
        console.debug(value);
    })
    .catch((error: Error)=>{
        console.debug(error.message);
    })
});
//console.log
// error1
// error1
//I'm expecting return1 then error1
```

johnbotris:
This allows you to stub an async function with a sequence of promises that may either resolve or reject.

It will also not allow you to provide non-promise values when stubbing async functions (which is good)

closed #478

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -63,6 +63,7 @@ export interface Stubber<D, R = D extends object ? Partial<D> : D> {
 }
 
 export interface PromiseStubber<P, R = P extends object ? Partial<P> : P> {
+  thenReturn<T>(first: Promise<R>, ...args: Array<Promise<R>>): TestDouble<T>;
   thenResolve<T>(first: R, ...args: Array<R>): TestDouble<T>;
   thenDo<T>(f: Function): TestDouble<T>;
   thenReject<T>(e: Error): TestDouble<T>;
--- a/test/safe/typescript-typings.test.ts
+++ b/test/safe/typescript-typings.test.ts
@@ -90,7 +90,7 @@ export = {
     td.when(f(td.matchers.not(false))).thenReject(new Error('rejected'))
 
     const asyncCat = td.instance(AsyncCat)
-    td.when(asyncCat.mewConcurrently()).thenReturn(Promise.resolve("purr"), Promise.reject("hiss!"))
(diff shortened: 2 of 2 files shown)


johnbotris:
@billtong would be great if you can check if it works for your use case, for mine it is good (I think we have the exact same situation)

johnbotris:
I will add some tests

searls:
Let me know when you think this is ready for review

billtong:
@johnbotris I just test my code with your commit. it's working great. Thanks!

johnbotris:
@billtong Thanks! 
@searls It's ready

PR #499, opened 2022-11-12
Improve replace typings

asdftd:


searls:
Seems like an improvement! As always with typing PRs I'm accepting this on trust and hope if it causes issues for anyone they'll let us know

asdftd:
Okay, thank you! Don't hesitate to ping me should there be any issues coming up.

searls:
landed in 3.16.7

PR #500, opened 2022-11-14
fix: replaceEsm typings

quinnturner:
Currently, as of the latest version 3.16.7: (#499, CC @asdftd @searls)

```ts
// `subscriptions` is `void`
const subscriptions = await td.replaceEsm(
      '../../subscriptions.js',
);
```

image

Now:

```ts
// `subscriptions` is `{ [namedExport: string]: any; default?: any; }`
const subscriptions = await td.replaceEsm(
      '../../subscriptions.js',
);
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -304,8 +304,8 @@ export function replace(obj: {}, property: string): TestDouble<any>;
  * @param {*} [defaultExportStub]
  * @returns {Promise<{default?: any, [namedExport: string]: any}>}
  */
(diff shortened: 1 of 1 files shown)


searls:
I take it to mean the ordering matters here?

quinnturner:
Yes, my description was unintentionally lazy.

The order of type definition overloading matters:
https://github.com/microsoft/TypeScript/issues/1860#issuecomment-72154737

searls:
Thanks! Fixed this in 3.16.8!

Issue #484, opened 2022-04-27
Support td.replace for Vitest


nullpaper:
How can we get Testdouble `replace` working with Vitest https://vitest.dev

Vitest and td almost work together smoothly. `td.replace` works fine when you're mocking a local object, but it doesn't work on an imported module:

```js
import * as td from 'test-double'

td.replace(Math, 'random', () => 1234)
Math.random()
// -> 1234 🎉

td.replace('some-module')
// -> Nope. Nothing replaced.
```

As Vitest is a 'drop-in' replacement for Jest, I tried using `testdouble-jest` but of course that doesn't work because Vitest is modules-based and Jest is commonjs so you end up with conflicts on `requireMock` rather than `importMock`. 

```
TypeError: Cannot set properties of undefined (setting 'requireMock')
> module.exports node_modules/testdouble-jest/index.js:25:23
```

But even tweaking these makes no difference as the import doesn't seem to get swapped out. I'm guessing this is a modules vs CJS issue somewhere in the deep. I'm hoping someone with more experience can point in the right direction.

### Environment

- testdouble 3.16.2
- testdouble-jest 2.0.0

rosston:
I've never heard of vitest before today, but FWIW, it says it includes [tinyspy](https://github.com/tinylibs/tinyspy), which says in its readme:
> Warning! Does not support ESM mocking. You can use `tinyspy` with `vitest`, who performs additional transformations to make ESM mocking work.

And from a quick look at the vitest docs, it looks like they have [a `mock` method](https://vitest.dev/api/#vi-mock) that seems similar to [Jest's](https://jestjs.io/docs/jest-object#jestmockmodulename-factory-options).

That error doesn't look like what I'd expect to see for a failure for ESM imports given [the source of testdouble-jest](https://github.com/testdouble/testdouble-jest/blob/e7ecc759d0aaeef6b7f007926ac1b7c261dbaab5/index.js#L25). @nullpaper Can you share how you're calling the `testdouble-jest` setup function as shown in [the installation section](https://github.com/testdouble/testdouble-jest#installation)?

searls:
I don't know what `vitest` is, but what's `test-double`?

```js
import * as td from 'test-double'
```

mcous:
I've been using testdouble for a long time and was also recently introduced to `vitest`. After noodling for a bit, I'm finding this to be a pretty serviceable integration of the two tools for my own needs:

```ts
import {vi, describe, it, afterEach} from 'vitest'
[...]

PR #501, opened 2022-11-23
docs: add testdouble-vitest links to docs

mcous:
Following up on https://github.com/testdouble/testdouble.js/issues/484#issuecomment-1323701359, this PR adds a couple links to [testdouble-vitest](https://github.com/mcous/testdouble-vitest) to the docs. This plugin adds `replaceEsm` support when using testdouble.js with [vitest](https://vitest.dev/) by way of vitest's built-in module replacement system, rather than through quibble.

I based it mostly off [testdouble-jest](https://github.com/testdouble/testdouble-jest), with changes to make it fit in with vitest - namely ES Modules and test API access via imports rather than globals.

I shuffled some links around in appendix A since it didn't really feel to me like my plugin (or testdouble-jest for that matter) were "Assertion Plugins", but please free to edit, revert, or whatever as you see fit!


searls:
Brilliant! Thank you!

PR #503, opened 2023-02-07
fix manual replacement in node-esm examples

Trim:
According to the current [documentation](https://github.com/testdouble/testdouble.js/blob/6535986b76e535edb9db8d1c029ca34335554ec5/docs/7-replacing-dependencies.md), replace signature use only two arguments to do manual replacement when first argument is a string.


Trim:
Oh, I've just seen there's tests for the examples and I've just broke them. Sorry, I'll check what is happening.

Trim:
I'm closing the request, because the example tests were working. It seems to override `default` method, the usage of `replace` require the three arguments.

PR #504, opened 2023-02-07
add documentation for td.replaceEsm API

Trim:
The API for `td.replaceEsm` differs a bit in argument interpretation compared to `td.replace` one, explain it and give documentation reference to `quibble.esm`.

I intentionally did not give details about the `default` possible value in `namedExportReplacements` because it requires a more recent version of quibble which has not been released yet.


searls:
Thanks for this!

PR #505, opened 2023-02-12
Add windows to the build

searls:
Ok @webstech, after merging in your quibble & teenytest changes into new releases and updating dependencies, I'm still struggling to get the build passing under windows. Could you take a look?


webstech:
> Could you take a look?

Have not had a chance to test yet but this appears to be the same as [this quibble change](https://github.com/testdouble/quibble/commit/1076467b87ec2df0c575783b3384a5255b65f2fa).  Caused by the npm differences between UNIXy systems and Windows.

webstech:
> Have not had a chance to test yet

I can confirm the windows test runs with that change, specifying `bash` on the `test:example:*` scripts in `package.json` and adding a `.gitattributes` file with '* eol=lf'.  Tested using cmd and wsl.

searls:
Looks good!

PR #506, opened 2023-02-13
ci: improve Windows support

webstech:
This patch series has been tested on Windows in cmd and wsl/bash environments.

1. ci - teenytest-proxy Windows support
    npm installs packages differently based on the platform.  The `.bin` entry the proxy was using is a symlink on UNIX type platforms but may have three different forms on WIndows.  The actual installed package will now  be required directly.  This is still a temporary patch pending fixes for the issues identified in the proxy file.

2. dev: support Windows based developers
    Add control so files are checked out with lf and not crlf.  A number of tests are shell scripts that run under bash and Windows lineends are not supported.

3. ci: do not expect shebang support on Windows
    Specify command to run in scripts named 'test:example:*'.

This is to support PR #505.


webstech:
@searls This PR is ready for your review and run.  I have verified it runs on GitHub actions windows-latest over [here](https://github.com/webstech/testteeny/actions/runs/4213505450/jobs/7313251483).

searls:
Does this end your 3-repo journey, @webstech? Or was there more to be done with respect to what semantic-release needs for windows support? Do I need to cut a new release of anything else for you?

webstech:
> Does this end your 3-repo journey, @webstech? Or was there more to be done with respect to what semantic-release needs for windows support? Do I need to cut a new release of anything else for you?

Sorry but your pain continues. ☹️I opened two more PRs to handle globing issues after chasing down a test case count difference.  It seems Windows prefers '"' quoting.  Some package.json used that and some used single quotes.  These changes are test related so new releases are not really needed.  

Thanks for bearing with me through this.  Windows support for semantic-release is available.  This journey started with getting the ci to run on Windows.  That is available now so will see if they add Windows to the ci runs.  This has been an interesting and educational experience.

searls:
All good, thanks for your help! The entire Node ecosystem will be win32 friendly by the time you're done

PR #472, opened 2021-10-04
Swap global for globalThis

NullVoxPopuli:
resolves
 - https://github.com/testdouble/testdouble.js/issues/471
 - https://github.com/testdouble/testdouble.js/issues/438

Issue #471, opened 2021-10-03
Switch global to globalThis


NullVoxPopuli:
### Description

This is a bit of a resurrection of: https://github.com/testdouble/testdouble.js/issues/438, in that browsers are not able to use testdouble.js without a `global` "polyfill", like the following:
```html
<script>
    // testdouble doesn't natively support the browser as it assumes
    // that a `global` identifier will be available, but that is not the case for browsers
    window.global = globalThis;
</script>
```

Just checking the compatibility table over here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis

it looks like globalThis is supported everywhere that matters? 

Is it possible to drop usage of `global`?

This is probably also related to: https://github.com/testdouble/testdouble.js/issues/434

### Environment

- [ ] `node -v` output:
- [ ] `npm -v` (or `yarn --version`) output:
- [ ] `npm ls testdouble` (or `yarn list testdouble`) version:

Chrome and Firefox v90+

importing testdouble causes an error, because of [this code](https://github.com/testdouble/testdouble.js/blob/65d7023b2f1c9084cdd7cd288550aa4ab68495bf/src/config.js#L5-L9): 

```
const DEFAULTS = {
  ignoreWarnings: false,
  promiseConstructor: global.Promise,
  suppressErrors: false
};
```

searls:
Seems good

rosston:
We might want to drop [support for Node 10](https://github.com/testdouble/testdouble.js/blob/7e0dc89a4ab899eb2e8b6ab91fe88011d88f0df6/.circleci/config.yml#L35) at the same time. If I'm tracking my Node changes correctly, v10 went EOL on 2021-04-30.

Issue #438, opened 2020-02-02
browser does not have `global` defined. Should this be `globalThis`?


NullVoxPopuli:
### Description

I want to load my tests. :)

### Issue
![image](https://user-images.githubusercontent.com/199018/73610955-f9725d00-45aa-11ea-8eec-c038bab9217f.png)

### Environment

- [ ] `node -v` output: Chrome (no node)
- [ ] `npm -v` (or `yarn --version`) output: (doesn't matter?)
- [x] `npm ls testdouble` (or `yarn list testdouble`) version: `testdouble@3.12.5`

I don't yet have time to do any digging, so I'm skiping the repro stuff below.
Just wanted to report this for now in case something obvious is happening.

I've been able to get passed this error with:

```html
    <script>
      window.global = window;
    </script>
```

but is this a hack? is this build-environment-dependent? idk

### Failing Test

- [ ] Fork the repo
- [ ] Add a failing test (probably to the `/regression/src' directory)
- [ ] Submit a pull request for the failing test or link to your branch here

### Example Repo

- [ ] Create a minimal repository that reproduces the issue
- [ ] Make sure that a fresh clone can run only `npm it` and observe the issue
- [ ] Link to that repo here

#### Runkit Notebook

- [ ] [Create a Runkit notebook](https://runkit.com/new)
- [ ] Invoke `var td = require('testdouble')` at the top
- [ ] Verify the behavior your issue is concerned with by clicking "Run"
- [ ] Link to the Runkit here

#### Code-fenced Examples

```js
var td = require('testdouble')

// Your steps here.
```

searls:
I spent a few hours on this and simply can't replicate it. Could you please provide a repo with which to reproduce this?

NullVoxPopuli:
How'd you try to reproduce it?

Just in the console, I reproduce the lack of `global`
![image](https://user-images.githubusercontent.com/199018/75585844-e034c300-5a27-11ea-806d-e2f33fb59242.png)

searls:
No I mean how are you installing and building your application such that you're seeing this behavior

NullVoxPopuli:
Sorry for my lack of understanding :( , but I don't understand the relevance? Do different environments polyfills different globals?

NullVoxPopuli:
To reproduce:

Clone repo and delete this line: https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/tests/index.html#L37

[...]

PR #508, opened 2023-03-12
Replace usage of global with globalThis for better browser support

maxbeatty:
I ran into the same issue as #438 and #471 but the workarounds didn't work. Our stack is TypeScript + esbuild + Karma + mocha (and hopefully testdouble 😀 ). Our library is completely browser-based so we only want to test in real browsers.

These changes are largely the same as #472 but also upgrades the supported node engine as mentioned in https://github.com/testdouble/testdouble.js/issues/471#issuecomment-933481803 . [Node.js Release Schedule](https://github.com/nodejs/release#release-schedule)

### DIFF
--- a/package.json
+++ b/package.json
@@ -148,6 +148,6 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">= 4.0.0"
+    "node": ">= 14"
   }
 }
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -148,6 +148,6 @@
   },
   "license": "MIT",
   "engines": {
-    "node": ">= 4.0.0"
+    "node": ">= 14"
```
searls, line 151: Literally laughed out loud at this. Good catch

searls:
Great patch, thank you! Landed in v3.17.0

PR #509, opened 2023-03-20
add missing function alias for ES module

Trim:
I'm using testdouble as an ES module (with `import * from 'testdouble'`) and I wasn't able to use the `td.function()`, but only the `td.func()` alias.

It seems the `index.mjs` forgot to export the name `function`.

This patch fix it by adding an alias for `func` in the `index.mjs` file.

I didn't find where to add a test for this, I'll do it if you can give me some insights.


searls:
Oops!

Fixed in testdouble@3.17.1

PR #511, opened 2023-03-28
make `thenThrow` and `thenReject` accept values of `any` type

iliocatallo:
Hello there,

[In TypeScript](https://github.com/microsoft/TypeScript/blob/f64f40d20511d2d2a9eb4fcc1ad6750977be2d40/src/lib/es2015.promise.d.ts#L42), `Promise.reject` accepts a reason of type `any`. 

Since `thenReject` is a short-hand for `thenReturn(Promise.reject)`, the proposal here is to make `thenReject` accept a parameter of type `any`. The same applies for `throw` vs. `thenThrow`.

Ciao!


searls:
Landed in v3.17.2
=

PR #515, opened 2023-05-05
feat(types): add "version" export to type declaration

jsejcksn:
I noticed this was missing in the type definitions.


searls:
thanks!

